PR #14, opened 2020-05-06
Add v1 and v2 machine names to access tokens

stephen-soltesz:
This change improves support for parsing machine names from mlab-ns and adds both v1 and v2 machine names to access tokens given to clients.

This change is necessary because the locate v2 API was designed to be v2-hostname aware from the beginning, but machines in the mlab-oti cluster still use v1 names. In practice today, this means that access tokens are rejected by the ndt-server because it is checking for a v1 hostname, when the token contains a v2 name.

Once mlab-ns natively returns v2 names, only the v2 name will be added to access tokens.

This change also adds two new flags to specify the legacy mlab-ns server address (currently defaulting to mlab-ns.appspot.com) and the platform project for v2 machine names (mlab-oti) which must be different than the runtime environment project (mlab-ns).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/14)


stephen-soltesz:
https://github.com/ooni/probe-engine/issues/562

stephen-soltesz:
https://github.com/m-lab/dev-tracker/issues/587

PR #24, opened 2020-07-23
Restore API key security

stephen-soltesz:
API key security was temporarily disabled pending complete HTTP referer patterns for one API key user. That is now complete. It should be safe to re-enable API key security, which this change does.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/24)

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -139,9 +139,8 @@ paths:
             request in the event of error.
           schema:
             $ref: "#/definitions/ErrorResult"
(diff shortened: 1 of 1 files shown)


stephen-soltesz:
The initial CB trigger failed, and I restarted it from the GCP console, which doesn't show up here unfortunately. But, it passed -- https://console.cloud.google.com/cloud-build/builds/69648df8-fff0-4dce-bfdd-2278257bca39

PR #37, opened 2020-10-29
Log the client protocol used to contact the server

stephen-soltesz:
Suspecting that some Locate v2 responses are being cached, it would be helpful to log the protocol used to connect to the server, i.e. http or https. Caching can be performed by third parties only when http requests are used, or when middleboxes intercept https connections (mitm, which we would not detect using this logging).

If we see a high rate of http requests, then we could consider adding an option to disallow that request type.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/37)

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -134,6 +134,8 @@ func (c *Client) TranslatedQuery(rw http.ResponseWriter, req *http.Request) {
 	// Set CORS policy to allow third-party websites to use returned resources.
 	rw.Header().Set("Content-Type", "application/json")
 	rw.Header().Set("Access-Control-Allow-Origin", "*")
+	// Prevent caching of result.
+	rw.Header().Set("Cache-Control", "private, max-age=0, no-transform")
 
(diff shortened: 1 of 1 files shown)


nkinkade (approved):
:lgtm:

Do you mean we may need to revisit this? https://github.com/m-lab/ndt-server/issues/217

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/37#-:-MKqNvCm0gD9M7RO1uTW:bbmq831)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade)

stephen-soltesz (commented):
I don't think so. I'm only concerned about transparent proxies caching responses from the Locate API. The communication between the client and ndt server need not be encrypted.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/37#-:-MKqOJbv7lsbQoDlz1iI:bz765kg)* status: :shipit: complete! 1 of 1 approvals obtained

Issue #40, opened 2020-11-06
User reports of Cloud VMs without geo location headers to Locate API


stephen-soltesz:
When the Locate API does not receive any location headers from App Engine, the request fails with 503.

The country fallback helped https://github.com/m-lab/locate/pull/29 , however we should consider adding a third-party geoip database as a final fallback, e.g. geolite2 used by uuid-annotator.

pboothe:
Consider adding "choose a random healthy server somewhere in the world" as a final fallback method.  If we can't geolocate the user, that seems like our problem not theirs, so returning an error code feels like giving up on a valid user request.

critzo:
If we do decide to use a third party service for this, I suggest that we consider options other than geolite2 as well. We should evaluate it's accuracy compared to other services like ipinfo.io, with whom we already have collaborated. I think we'll find that it is much more accurate than maxmind.

stephen-soltesz:
@pboothe there appear to be two default positions to this situation.

1. "Clearly" the server should return random results if it cannot make a better determination
2. "Clearly" random results from the server are not what a user or M-Lab want

I brought this question to the team at one of the engineering discussion meetings. No one spoke in defense of the random response. The rationale being:

* A random target server will not perform well or consistently
* A random target server will not provide a user with helpful information about their connectivity
* A random target server will not provide helpful measurement data for M-Lab
* A random target server masks an underlying issue
* A random target server is not consistent with the Locate API's "/v2/nearest" query

In this case, a user reported the problem because it wasn't working at all. And, we can add a better strategy. It's not clear to me that this would have happened with random results.

stephen-soltesz:
@critzo There is already support using geolite2 in the queue. However, we're creating an architecture that would make it very simple to add other sources too, like ipinfo.io. In general, it would be preferable perform target selection using the same client geo as we use to annotate the data. There's a lot here of course. So, I'm not saying anything is set in stone.

stephen-soltesz:
https://github.com/m-lab/locate/releases/tag/v0.7.0

PR #42, opened 2020-11-18
Add clientgeo package for appengine location

stephen-soltesz:
This change moves the "findLocation" logic from the http handler to a new `clientgeo` package. We want to add additional methods of looking up the client location. This is the first step to making an general interface for that operation.

This change better distinguishes the ServiceUnavailable and InternalServiceError cases. Unit tests are updated accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/42)

### DIFF
--- a/clientgeo/appengine.go
+++ b/clientgeo/appengine.go
@@ -1,0 +1,103 @@
+package clientgeo
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"strings"
+
+	"github.com/apex/log"
+	"github.com/m-lab/locate/static"
+)
+
+var (
+	// ErrBadLatLonFormat is returned with a lat,lon header is missing or corrupt.
+	ErrBadLatLonFormat = errors.New("lat,lon format was missing or corrupt")
+
+	// ErrNullLatLon is returned with a 0,0 lat/lon value is provided.
+	ErrNullLatLon = errors.New("lat,lon value was null: " + nullLatLon)
+
+	latlonMethod  = "appengine-latlong"
+	regionMethod  = "appengine-region"
+	countryMethod = "appengine-country"
+	noneMethod    = "appengine-none"
+	nullLatLon    = "0.000000,0.000000"
+)
+
+// NewAppEngineLocator thing.
+func NewAppEngineLocator() *AppEngineLocator {
+	return &AppEngineLocator{}
+}
+
+// AppEngineLocator thing.
+type AppEngineLocator struct{}
+
+// Locate finds a location for the given client request.
+func (sl *AppEngineLocator) Locate(req *http.Request) (*Location, error) {
+	headers := req.Header
+	fields := log.Fields{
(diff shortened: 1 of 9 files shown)


stephen-soltesz:
Part of https://github.com/m-lab/locate/issues/40

pboothe (approved):
This looks good to me. I was about to ask you to do the rest of #40 here, but this is a good step towards that and I'm glad to see that adding the fallback of MaxMind or something like it is part of the plan. 

One request for monitoring, but it looks like the appengine monitoring story is likely taken care of, so feel free to just say "yup" and then submit. :)

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/42#-:-MMUlpSEB7G70v6bPExb:bdz6fe7)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pboothe and @stephen-soltesz)

___
*[handler/handler.go, line 115 at r1](https://reviewable.io/reviews/m-lab/locate/42#-MMUfswI6AhRQG7JVkBo:-MMUfswI6AhRQG7JVkBp:b-14jbo0) ([raw file](https://github.com/m-lab/locate/blob/1b7f319111c15b55866eaa6ceb3422bc0f4ff65d/handler/handler.go#L115)):*
> ```Go
> 	if err != nil {
> 		status := http.StatusServiceUnavailable
> 		result.Error = v2.NewError("nearest", "Failed to lookup nearest machines", status)
> ```

I feel like we should increment a metric here? Or somewhere? I guess AppEngine isn't monitored that way. But I would like to be sure we can easily track how many times this happens.

stephen-soltesz (commented):
Thank you! The rest of #40 is about the same size as this PR. Combined it was too much for a single PR, imo. I'll have another PR soon.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/42#-:-MMWqH6E3RBmZ7cQRvd1:b-55mp3m)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pboothe)

___
*[handler/handler.go, line 115 at r1](https://reviewable.io/reviews/m-lab/locate/42#-MMUfswI6AhRQG7JVkBo:-MMWqedX5NdJwXcsk73i:b8mzpib) ([raw file](https://github.com/m-lab/locate/blob/1b7f319111c15b55866eaa6ceb3422bc0f4ff65d/handler/handler.go#L115)):*
<details><summary><i>Previously, pboothe (Peter Boothe) wrote…</i></summary><blockquote>

I feel like we should increment a metric here? Or somewhere? I guess AppEngine isn't monitored that way. But I would like to be sure we can easily track how many times this happens.
</blockquote></details>

We monitor and alert on AE status codes, 5xx in particular. https://github.com/m-lab/prometheus-support/blob/master/config/federation/prometheus/alerts.yml#L377

That addresses 500 errors from either mlab-ns or the locate v2 API. These are rather rare.

IMO, this change make the semantics of "ServiceUnavailable" clearer than it was before.

There is still room for improvement. If this does not address your immediate concerns, please lmk.

PR #45, opened 2020-12-01
Use constant names for X-Locate-Client* headers

stephen-soltesz:
This change replaces all instances of "X-Locate-Client*" headers with constant names to fix a bug where the maxmind headers were not returned correctly.

This bug was caused by the inline definition of `http.Headers` with non-canonical header key names, https://github.com/m-lab/locate/blob/e2fd31484cdb6603f4209cfe294b23f965d4c069/clientgeo/maxmind.go#L78-L81

combined with the header copy loop using `http.Header.Get` and `http.Header.Set` methods, which normalize key names strictly using `textproto.CanonicalMIMEHeaderKey`. https://github.com/m-lab/locate/blob/e2fd31484cdb6603f4209cfe294b23f965d4c069/handler/handler.go#L120-L122

So, the copy loop used canonical names while the underlying map had non-canonical keys. As a result, they were unnameable.

This bug does not effect the maxmind location lookup, only the reporting of location&method to the caller.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/45)

### DIFF
--- a/clientgeo/appengine.go
+++ b/clientgeo/appengine.go
@@ -50,8 +50,8 @@ func (sl *AppEngineLocator) Locate(req *http.Request) (*Location, error) {
 	loc, err := splitLatLon(latlon)
 	if err == nil {
 		log.WithFields(fields).Info(latlonMethod)
-		loc.Headers.Set("X-Locate-ClientLatLon", latlon)
-		loc.Headers.Set("X-Locate-ClientLatLon-Method", latlonMethod)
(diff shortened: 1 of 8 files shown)


pboothe (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/45#-:-MNV0nf_1dVjoYqUQ-Pl:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @pboothe)

PR #54, opened 2021-07-21
Adds new package named secrets

nkinkade:
This PR adds a new package named `secrets`. It serves the same general purpose as the current `decrypt` package, but uses a different backing mechanism. Instead of using the GCP Key Management Service (KMS) (like `decrypt`), this new package uses the GCP Secret Manager (SM).

The primary motivation for moving to the SM is that, currently, the signer and verify keys for the Locate Service are stored in this public repository in encrypted form. It is probably safe, but perhaps better to remove them altogether in any form. Moving to the SM allows us to remove the keys 100% from this repo. This new module will allow the Locate Service to fetch the keys directly from the SM at runtime.

Another advantage of using the SM over KMS, is that the SM natively supports secret versions, which will allow us to more easily rotate the keys in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/54)

### DIFF
--- a/secrets/load.go
+++ b/secrets/load.go
@@ -57,7 +57,7 @@ func (c *Config) getSecret(ctx context.Context, client SecretClient, path string
 
 	result, err := client.AccessSecretVersion(ctx, req)
 	if err != nil {
-		return []byte{}, err
+		return nil, err
 	}
 
 	return result.Payload.Data, nil
--- a/secrets/load.go
+++ b/secrets/load.go
@@ -1,4 +1,4 @@
-// Package secrets loads secrets from the Google Cloud Secret Manager
+// Package secrets loads secrets from the Google Cloud Secret Manager.
 package secrets
 
 import (
--- a/secrets/load.go
+++ b/secrets/load.go
@@ -1,0 +1,131 @@
+// Package secrets loads secrets from the Google Cloud Secret Manager
+package secrets
+
+import (
+	"context"
+	"fmt"
+
+	secretmanager "cloud.google.com/go/secretmanager/apiv1"
+	"github.com/googleapis/gax-go"
+	"github.com/m-lab/access/token"
+	"google.golang.org/api/iterator"
+	secretmanagerpb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1"
+)
+
+// SecretClient wraps the AccessSecretVersion function provided by the
+// secretmanager.Client.
+type SecretClient interface {
+	AccessSecretVersion(ctx context.Context, req *secretmanagerpb.AccessSecretVersionRequest, opts ...gax.CallOption) (*secretmanagerpb.AccessSecretVersionResponse, error)
+	ListSecretVersions(ctx context.Context, req *secretmanagerpb.ListSecretVersionsRequest, opts ...gax.CallOption) *secretmanager.SecretVersionIterator
+}
+
+// iter warps the Next() method of a *secretmanager.SecretVersionIterator.
+type iter interface {
+	Next(it *secretmanager.SecretVersionIterator) (*secretmanagerpb.SecretVersion, error)
+}
+
+// stdIter implements the iter interfaces, and is used to invoke the
+// iterator.Next() method.
+type stdIter struct{}
+
+// Next invokes the Next() method of a *secretmanager.SecretVersionIterator.
+func (s *stdIter) Next(it *secretmanager.SecretVersionIterator) (*secretmanagerpb.SecretVersion, error) {
+	return it.Next()
+}
+
+// Config contains settings for secrets.
+type Config struct {
+	iter    iter
+	Name    string
+	Project string
+}
+
+// NewConfig creates a new secret config.
+func NewConfig(project string) *Config {
+	return &Config{
+		iter:    &stdIter{},
+		Project: project,
+	}
+}
+
+// getSecret fetches the version of a secret specified by 'path' from the Secret
+// Manager API.
+func (c *Config) getSecret(ctx context.Context, client SecretClient, path string) ([]byte, error) {
+	req := &secretmanagerpb.AccessSecretVersionRequest{
+		Name: path,
+	}
+
+	result, err := client.AccessSecretVersion(ctx, req)
+	if err != nil {
+		return []byte{}, err
+	}
+
+	return result.Payload.Data, nil
+}
+
+// getSecretVersions returns a slice of all *enabled* versions for a secret. It
+// will ignore disabled for destroyed versions of a secret.
+func (c *Config) getSecretVersions(ctx context.Context, client SecretClient) ([]string, error) {
+	req := &secretmanagerpb.ListSecretVersionsRequest{
+		Parent:   c.path(),
+		PageSize: 1000,
+	}
+
+	it := client.ListSecretVersions(ctx, req)
+	versions := []string{}
+	for {
+		resp, err := c.iter.Next(it)
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		if resp.State != secretmanagerpb.SecretVersion_ENABLED {
+			continue
+		}
+		versions = append(versions, resp.Name)
+	}
+
+	if len(versions) < 1 {
+		return nil, fmt.Errorf("no versions found for secret: %s", c.Name)
+	}
+
+	return versions, nil
+}
+
+// LoadSigner fetches the latest version of the named secret containing the JWT
+// signer key from the Secret Manager API and returns a *token.Signer.
+func (c *Config) LoadSigner(ctx context.Context, client SecretClient) (*token.Signer, error) {
+	versions, err := c.getSecretVersions(ctx, client)
+	if err != nil {
+		return nil, err
+	}
+	key, err := c.getSecret(ctx, client, versions[0])
+	if err != nil {
+		return nil, err
+	}
+	return token.NewSigner(key)
+}
+
+// LoadVerifier fetches all enabled versions of the named secret containing the
+// JWT verifier keys and returns a * token.Verifier.
+func (c *Config) LoadVerifier(ctx context.Context, client SecretClient) (*token.Verifier, error) {
+	versions, err := c.getSecretVersions(ctx, client)
+	if err != nil {
+		return nil, err
+	}
+	keys := [][]byte{}
+	for _, version := range versions {
+		key, err := c.getSecret(ctx, client, version)
+		if err != nil {
+			return nil, err
+		}
+		keys = append(keys, key)
+	}
+	return token.NewVerifier(keys...)
+}
+
+func (c *Config) path() string {
+	return "projects/" + c.Project + "/secrets/" + c.Name
+}
--- a/secrets/load_test.go
+++ b/secrets/load_test.go
@@ -74,24 +74,36 @@ func Test_getSecret(t *testing.T) {
 
 	var secretData [][]byte
 	secretData = append(secretData, []byte("fake-secret"))
-	client := &fakeSecretClient{
-		data:    secretData,
-		wantErr: false,
+
+	tests := []struct {
+		wantErr bool
+	}{
+		{
+			wantErr: false,
+		},
+		{
+			wantErr: true,
+		},
 	}
 
-	secret, err := cfg.getSecret(ctx, client, "fake-path")
-	if err != nil {
-		t.Fatalf("Did not expect an error, but got error: %s", err)
-	}
-	if string(secret) != string(secretData[0]) {
-		t.Fatalf("Expected secret value '%s', but got: %s", string(secretData[0]), string(secret))
-	}
+	for _, tt := range tests {
+		client := &fakeSecretClient{
+			data:    secretData,
+			wantErr: tt.wantErr,
+		}
 
-	client.wantErr = true
+		secret, err := cfg.getSecret(ctx, client, "fake-path")
 
-	_, err = cfg.getSecret(ctx, client, "fake-path")
-	if err == nil {
-		t.Fatal("Wanted error, but did not get one.")
+		if (err != nil) != tt.wantErr {
+			t.Fatalf("Got error: %v, but wantErr is %v", err, tt.wantErr)
+			return
+		}
+
+		if !tt.wantErr {
+			if string(secret) != string(secretData[0]) {
+				t.Fatalf("Expected secret value '%s', but got: %s", string(secretData[0]), string(secret))
+			}
+		}
 	}
 }
 
@@ -160,7 +172,8 @@ func Test_getSecretVersions(t *testing.T) {
 		versions, err := cfg.getSecretVersions(ctx, client)
 
 		if (err != nil) != tt.wantErr {
-			t.Fatalf("Did not expect error, but got error: %s", err)
+			t.Fatalf("Got error: %v, but wantErr is %v", err, tt.wantErr)
+			return
 		}
 
 		if len(versions) != tt.expectedCount {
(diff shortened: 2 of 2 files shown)


stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/54)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[secrets/load.go, line 1 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfDf4yp4YvzwRpQNgw9:-MfDf4yp4YvzwRpQNgwA:bdbhfxg) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load.go#L1)):*
> ```Go
> // Package secrets loads secrets from the Google Cloud Secret Manager
> ```

nit: please end sentences with period.
___
*[secrets/load.go, line 60 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfDg8DQEzjEesL5kNCo:-MfDg8DQEzjEesL5kNCp:b-47jvmr) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load.go#L60)):*
> ```Go
> 	result, err := client.AccessSecretVersion(ctx, req)
> 	if err != nil {
> 		return []byte{}, err
> ```

Prefer returning `nil` on error.
___
*[secrets/load_test.go, line 22 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcVhm01HYTGBuLHpRh:-MfEdtsgBtAYEm1Sm1cU:bnagh02) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L22)):*
> ```Go
> func (f *fakeSecretClient) AccessSecretVersion(ctx context.Context, req *secretmanagerpb.AccessSecretVersionRequest, opts ...gax.CallOption) (*secretmanagerpb.AccessSecretVersionResponse, error) {
> 	defer f.incrementIdx()
> 	if !f.wantErr {
> ```

It looks like only one condition is necessary. Check one initially and let the rest of the function logic be unconditional.

For example, begin by checking `if f.wantErr`, then remove this condition.

Test logic must be correct by inspection. In general, try to make it as simple as possible (but not simpler :) )
___
*[secrets/load_test.go, line 35 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcW8S84SSk9uPrRVT0:-MfcW8S84SSk9uPrRVT1:b-klod71) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L35)):*
> ```Go
> 	}
> 
> 	return &secretmanagerpb.AccessSecretVersionResponse{}, nil
> ```

Is this reachable? `wantErr` is either true or false. There are three return paths.
___
*[secrets/load_test.go, line 49 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWOSK-MTSd1jpc8ch:-MfcWOSK-MTSd1jpc8ci:bcryvtb) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L49)):*
> ```Go
> 	idx      int
> 	versions []*secretmanagerpb.SecretVersion
> 	err      string
> ```

Typically a variable named `err` will contain an `Error` type.

In cases like this, I've added fields like:

```
type fakeIter struct {
	idx      int
	versions []*secretmanagerpb.SecretVersion
	wantFooErr bool
        wantBarErr bool
```

And, then check for those explicit cases. The default case is "success" or "typical" (whatever that means for the situation).

This makes the error cases that the fake handles a little self-describing.
___
*[secrets/load_test.go, line 55 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWptP3-dmH5uW_jzm:-MfcWptP3-dmH5uW_jzn:b-camq25) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L55)):*
> ```Go
> 	defer f.incrementIdx()
> 
> 	if f.err == "no-errors" {
> ```

This looks like a success case?
___
*[secrets/load_test.go, line 65 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWzHm6oXP-r8gJBvj:-MfcWzHm6oXP-r8gJBvk:b-j85c1y) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L65)):*
> ```Go
> 	}
> 
> 	if f.err == "no-versions" {
> ```

What is the difference between the "no-errors" case and the "no-versions" case? (it may be subtle -- I can't see it quickly).

nkinkade (commented):
Thanks for the feedback. I believe I have addressed your comments. PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/54#-:-Mfcl_Uz0buLfGyeZkNg:bizuke1)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[secrets/load.go, line 1 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfDf4yp4YvzwRpQNgw9:-MfcbflR1KfDKPzbF3Sl:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load.go#L1)):*

nit: please end sentences with period.
</blockquote></details>

Done.
___
*[secrets/load.go, line 60 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfDg8DQEzjEesL5kNCo:-MfcbwLw2cEsi_ycizy4:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load.go#L60)):*

Prefer returning `nil` on error.
</blockquote></details>

Done.
___
*[secrets/load_test.go, line 22 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcVhm01HYTGBuLHpRh:-MfccCmj8tFEmwQBXKSF:b-x10cnv) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L22)):*

It looks like only one condition is necessary. Check one initially and let the rest of the function logic be unconditional.

For example, begin by checking `if f.wantErr`, then remove this condition.

Test logic must be correct by inspection. In general, try to make it as simple as possible (but not simpler :) )
</blockquote></details>

Done. Thanks, this definitely simplifies the function.
___
*[secrets/load_test.go, line 35 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcW8S84SSk9uPrRVT0:-MfccWC84-j-b_cFLu7f:bdxp1pv) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L35)):*

Is this reachable? `wantErr` is either true or false. There are three return paths.
</blockquote></details>

It was not, but it is now removed as part of the improvement in your suggestion above about removing the unneeded conditional.
___
*[secrets/load_test.go, line 49 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWOSK-MTSd1jpc8ch:-MfckLdX2krjAm_zb6wS:bn8faht) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L49)):*

Typically a variable named `err` will contain an `Error` type.

In cases like this, I've added fields like:

```
type fakeIter struct {
	idx      int
	versions []*secretmanagerpb.SecretVersion
	wantFooErr bool
        wantBarErr bool
```

And, then check for those explicit cases. The default case is "success" or "typical" (whatever that means for the situation).

This makes the error cases that the fake handles a little self-describing.
</blockquote></details>

I have made updates here to make things more clear, as well as other places in the file that were using similar unclear logic and names.
___
*[secrets/load_test.go, line 55 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWptP3-dmH5uW_jzm:-MfckVB27XflTdHCxj4i:b-fcxfu3) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L55)):*

This looks like a success case?
</blockquote></details>

It is. This is should be clearer now, based on changes from your feedback in the previous comment.
___
*[secrets/load_test.go, line 65 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWzHm6oXP-r8gJBvj:-Mfckbxa3XbwfjhYMgik:b-6d7qyd) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L65)):*

What is the difference between the "no-errors" case and the "no-versions" case? (it may be subtle -- I can't see it quickly). 
</blockquote></details>

I have renamed the "no-errors" case to "success" for clarity. The "no-versions" case is an error, because if no versions of a secret can be found in Secret Manager, then we have no JWT keys to load on initialization.

stephen-soltesz (changes requested):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/54)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[secrets/load_test.go, line 49 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWOSK-MTSd1jpc8ch:-MfdBXxm42HwLc_4RlwG:b-amm6b6) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L49)):*

I have made updates here to make things more clear, as well as other places in the file that were using similar unclear logic and names.
</blockquote></details>

`wantSuccess` should be the default case, so need not be specified; when all the wantErr cases are False.
___
*[secrets/load_test.go, line 65 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWzHm6oXP-r8gJBvj:-MfdAapo3rigUX2s89kO:b-qm33be) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L65)):*

I have renamed the "no-errors" case to "success" for clarity. The "no-versions" case is an error, because if no versions of a secret can be found in Secret Manager, then we have no JWT keys to load on initialization.
</blockquote></details>

I still don't see a logic difference. Are the two cases with the same logic? If the logic is the same, I believe the code will be simpler with a single condition.

nkinkade (commented):
@stephen-soltesz: I believe I have addressed your comments. PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/54#-:-MfdRFn10KawWQuqlkpl:bcbpk2l)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[secrets/load_test.go, line 49 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWOSK-MTSd1jpc8ch:-MfdOz7zD-Lp3PkaSjol:bkkd4c4) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L49)):*

`wantSuccess` should be the default case, so need not be specified; when all the wantErr cases are False.
</blockquote></details>

Fixed, I believe.
___
*[secrets/load_test.go, line 65 at r1](https://reviewable.io/reviews/m-lab/locate/54#-MfcWzHm6oXP-r8gJBvj:-MfdQtznDnDU_hBgxeON:b-j0m5uw) ([raw file](https://github.com/m-lab/locate/blob/edb7b9e15829c589d838b6d7a98e96d30daa68d2/secrets/load_test.go#L65)):*

I still don't see a logic difference. Are the two cases with the same logic? If the logic is the same, I believe the code will be simpler with a single condition.
</blockquote></details>

I see what you mean now. Fixed, I believe.

stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/54)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[secrets/load_test.go, line 21 at r3](https://reviewable.io/reviews/m-lab/locate/54#-MfdtKEf-fdrmS1tGM0j:-MfdtKEf-fdrmS1tGM0k:b-snqxiw) ([raw file](https://github.com/m-lab/locate/blob/48dc53068ea4fe87d0878dd37ebae192f8c7f5df/secrets/load_test.go#L21)):*
> ```Go
> 
> func (f *fakeSecretClient) AccessSecretVersion(ctx context.Context, req *secretmanagerpb.AccessSecretVersionRequest, opts ...gax.CallOption) (*secretmanagerpb.AccessSecretVersionResponse, error) {
> 	defer f.incrementIdx()
> ```

Should this increment if `wantErr` is true?
___
*[secrets/load_test.go, line 82 at r3](https://reviewable.io/reviews/m-lab/locate/54#-MfdxfGlBjeYHG0H4axA:-MfdxfGlBjeYHG0H4axB:b3qruh0) ([raw file](https://github.com/m-lab/locate/blob/48dc53068ea4fe87d0878dd37ebae192f8c7f5df/secrets/load_test.go#L82)):*
> ```Go
> 	}
> 
> 	secret, err := cfg.getSecret(ctx, client, "fake-path")
> ```

I understand that sometimes table driven tests aren't as helpful as explicit tests. The other tests here use table-driven testing. Why is this one different?  If there's no particular reason, I recommend preferring the table driven approach here also. If there is, please add a comment.

nkinkade (commented):
@stephen-soltesz: Thank you for detailed review. PTAL?

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/54#-:-MfeAN7y-DNrKyEXt7Dr:bcowflb)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[secrets/load_test.go, line 21 at r3](https://reviewable.io/reviews/m-lab/locate/54#-MfdtKEf-fdrmS1tGM0j:-Mfe0pIy-6ZaCK_I44xg:bwv90vt) ([raw file](https://github.com/m-lab/locate/blob/48dc53068ea4fe87d0878dd37ebae192f8c7f5df/secrets/load_test.go#L21)):*

Should this increment if `wantErr` is true?
</blockquote></details>

Good catch. Your question made me realize that one of my test cases was getting an error for the wrong reason. The second test in ` Test_getSecretVersions()` wants an error not for an iterator error, but because no valid versions were returned. I believe I have fixed this. And I moved the `defer` statement after the error case here, and did the same for `fakeIter.Next()`.
___
*[secrets/load_test.go, line 82 at r3](https://reviewable.io/reviews/m-lab/locate/54#-MfdxfGlBjeYHG0H4axA:-Mfe9rjw5ixXE5s28L1-:b-nx4b07) ([raw file](https://github.com/m-lab/locate/blob/48dc53068ea4fe87d0878dd37ebae192f8c7f5df/secrets/load_test.go#L82)):*

I understand that sometimes table driven tests aren't as helpful as explicit tests. The other tests here use table-driven testing. Why is this one different?  If there's no particular reason, I recommend preferring the table driven approach here also. If there is, please add a comment.
</blockquote></details>

No particular reason. Maybe I felt like the testing was so simple and short that it was more work to work up the table-driven testing than to just handle each of the two tests manually. I've converted the test to be table-driven, and it is probably a bit cleaner this way.

I also discovered that the error message for when the value of `err` was contrary to the value of `wantErr` was incorrect in all the tests, so I fixed that too.

stephen-soltesz (approved):
:lgtm: - Thank you!

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/54#-:-MfhSQqN4KUD-IjIY3Sw:bsjzsje)* status: :shipit: complete! 1 of 1 approvals obtained

PR #57, opened 2021-10-12
Increases max instances to 20 for Locate Service

nkinkade:
There is no need to limit the number of instances of the Locate Service
that AppEngine can launch to scale up the app. The default max is 8, and
this change sets the max to what I hope will be a very generous ceiling
of 20. I removed the min number of instances configurations, since its
value (2) was equal to the default for automatic scaling.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/57)

### DIFF
--- a/cmd/monitoring-token/monitoring-token.go
+++ b/cmd/monitoring-token/monitoring-token.go
@@ -18,7 +18,7 @@ import (
 	v2 "github.com/m-lab/locate/api/v2"
 	"github.com/m-lab/locate/proxy"
 	"github.com/m-lab/locate/static"
-	"gopkg.in/go-jose/go-jose.v2/jwt"
+	"gopkg.in/square/go-jose.v2/jwt"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/57)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[-- commits, line 20 at r1](https://reviewable.io/reviews/m-lab/locate/57#-MltSrATAOzVlKP8gUEu):*

Oooooh. The URL / redirect that Go was seeing may be different than what my browser was seeing. I'm guessing the old one had a bad cert?
</blockquote></details>

I was able to `go get gopkg.in/go-jose/go-jose.v2/jwt` on my local machine with no issues. It was only in the Travis build env where it was failing. Reinstalling the ca-certificates package _before_ doing the build fixed the issue. But I also reverted this commit to fetch the package at location `gopkg.in/square/go-jose.v2/jwt`, since using the other one was causing build failures. I expected them to be the same code, in two different repos, but that was not apparently the case.

PR #64, opened 2022-04-27
Add heartbeat service connection

cristinaleonr:
This change adds the heartbeat service.

It uses the connection package to establish a persistent connection with the Locate and send heartbeat messages at periodic intervals. If the connection is dropped, it automatically tries to reconnect using an exponential backoff implementation for retries. 

This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/64)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,9 @@ script:
 - go vet ./...
 - go test ./... -cover=1 -coverprofile=_c.cov
 - go test ./... -race
+
+after_success:
+# Note: Do this in the after_success stage so that
+# a broken coveralls won't cause the build to fail.
+# (details: https://github.com/lemurheavy/coveralls-public/issues/1264)
 - $GOPATH/bin/goveralls -service=travis-ci -coverprofile=_c.cov
--- a/cmd/heartbeat/README.md
+++ b/cmd/heartbeat/README.md
@@ -1,18 +1,19 @@
 # Heartbeat Service
 
 The Heartbeat Service is a sidecar service that performs health checks
-for experiment instances running in the same Kubernetes pod.
+for experiment instances running in the same Kubernetes pod and reports
+them to the Locate API.
 
 On start, it establishes a WebSocket connection with the Locate Service
 and sends an initial registration message. Subsequently, it performs
-periodic instance health checks and sends the results to the Locate.
+periodic instance health checks and sends the results to the Locate API.
 
 ## Local Development
 
 To run the service locally, build the package and pass a test URL
-in the `-locate-url` flag.
+in the `-heartbeat-url` flag.
 
 ```sh
 $ go build
-$ ./heartbeat -locate-url=ws://locate-dot-mlab-sandbox.appspot.com/v2/platform/heartbeat/
+$ ./heartbeat -heartbeat-url=ws://locate-dot-mlab-sandbox.appspot.com/v2/platform/heartbeat/
 ```
--- a/cmd/heartbeat/README.md
+++ b/cmd/heartbeat/README.md
@@ -1,0 +1,18 @@
+# Heartbeat Service
+
+The Heartbeat Service is a sidecar service that performs health checks
+for experiment instances running in the same Kubernetes pod.
+
+On start, it establishes a WebSocket connection with the Locate Service
+and sends an initial registration message. Subsequently, it performs
+periodic instance health checks and sends the results to the Locate.
+
+## Local Development
+
+To run the service locally, build the package and pass a test URL
+in the `-locate-url` flag.
+
+```sh
+$ go build
+$ ./heartbeat -locate-url=ws://locate-dot-mlab-sandbox.appspot.com/v2/platform/heartbeat/
+```
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -22,7 +22,7 @@ func init() {
 
 func main() {
 	flag.Parse()
-	rtx.Must(flagx.ArgsFromEnvWithLog(flag.CommandLine, false), "Failed to read args from env")
+	rtx.Must(flagx.ArgsFromEnvWithLog(flag.CommandLine, false), "failed to read args from env")
 
 	conn := connection.NewConn()
 	conn.Dial(locate, http.Header{})
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -15,13 +15,13 @@ import (
 )
 
 var (
-	locate              string
+	heartbeatURL        string
 	heartbeatPeriod     = static.HeartbeatPeriod
 	mainCtx, mainCancel = context.WithCancel(context.Background())
 )
 
 func init() {
-	flag.StringVar(&locate, "locate-url", "ws://localhost:8080/v2/platform/heartbeat/",
+	flag.StringVar(&heartbeatURL, "heartbeat-url", "ws://localhost:8080/v2/platform/heartbeat/",
 		"URL for locate service")
 }
 
@@ -30,8 +30,8 @@ func main() {
 	rtx.Must(flagx.ArgsFromEnvWithLog(flag.CommandLine, false), "failed to read args from env")
 
 	conn := connection.NewConn()
-	err := conn.Dial(locate, http.Header{})
-	rtx.Must(err, "failed to establish a websocket connection with %s", locate)
+	err := conn.Dial(heartbeatURL, http.Header{})
+	rtx.Must(err, "failed to establish a websocket connection with %s", heartbeatURL)
 
 	write(conn)
 }
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"log"
 	"net/http"
@@ -14,8 +15,9 @@ import (
 )
 
 var (
-	locate string
-	done   = make(chan bool, 1)
+	locate              string
+	heartbeatPeriod = static.HeartbeatPeriod
+	mainCtx, mainCancel = context.WithCancel(context.Background())
 )
 
 func init() {
@@ -38,12 +40,12 @@ func main() {
 // HeartbeatPeriod.
 func write(ws *connection.Conn) {
 	defer ws.Close()
-	ticker := *time.NewTicker(static.HeartbeatPeriod)
+	ticker := *time.NewTicker(heartbeatPeriod)
 	defer ticker.Stop()
 
 	for {
 		select {
-		case <-done:
+		case <-mainCtx.Done():
 			return
 		case <-ticker.C:
 			err := ws.WriteMessage(websocket.TextMessage, []byte("Health message!"))
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -13,7 +13,10 @@ import (
 	"github.com/m-lab/locate/static"
 )
 
-var locate string
+var (
+	locate string
+	done   = make(chan bool, 1)
+)
 
 func init() {
 	flag.StringVar(&locate, "locate-url", "ws://localhost:8080/v2/platform/heartbeat/",
@@ -25,7 +28,9 @@ func main() {
 	rtx.Must(flagx.ArgsFromEnvWithLog(flag.CommandLine, false), "failed to read args from env")
 
 	conn := connection.NewConn()
-	conn.Dial(locate, http.Header{})
+	err := conn.Dial(locate, http.Header{})
+	rtx.Must(err, "failed to establish a websocket connection with %s", locate)
+
 	write(conn)
 }
 
@@ -33,14 +38,18 @@ func main() {
 // HeartbeatPeriod.
 func write(ws *connection.Conn) {
 	defer ws.Close()
-	ticker := time.NewTicker(static.HeartbeatPeriod)
+	ticker := *time.NewTicker(static.HeartbeatPeriod)
 	defer ticker.Stop()
 
 	for {
-		<-ticker.C
-		err := ws.WriteMessage(websocket.TextMessage, []byte("Health message!"))
-		if err != nil {
-			log.Printf("failed to write, err: %v", err)
+		select {
+		case <-done:
+			return
+		case <-ticker.C:
+			err := ws.WriteMessage(websocket.TextMessage, []byte("Health message!"))
+			if err != nil {
+				log.Printf("failed to write, err: %v", err)
+			}
 		}
 	}
 }
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -16,6 +16,7 @@ import (
 
 var (
 	locate              string
+	heartbeatPeriod     = static.HeartbeatPeriod
 	mainCtx, mainCancel = context.WithCancel(context.Background())
 )
 
@@ -39,7 +40,7 @@ func main() {
 // HeartbeatPeriod.
 func write(ws *connection.Conn) {
 	defer ws.Close()
-	ticker := *time.NewTicker(static.HeartbeatPeriod)
+	ticker := *time.NewTicker(heartbeatPeriod)
 	defer ticker.Stop()
 
 	for {
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -16,7 +16,6 @@ import (
 
 var (
 	locate              string
-	heartbeatPeriod = static.HeartbeatPeriod
 	mainCtx, mainCancel = context.WithCancel(context.Background())
 )
 
@@ -40,7 +39,7 @@ func main() {
 // HeartbeatPeriod.
 func write(ws *connection.Conn) {
 	defer ws.Close()
-	ticker := *time.NewTicker(heartbeatPeriod)
+	ticker := *time.NewTicker(static.HeartbeatPeriod)
 	defer ticker.Stop()
 
 	for {
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"flag"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/gorilla/websocket"
+	"github.com/m-lab/go/flagx"
+	"github.com/m-lab/go/rtx"
+	"github.com/m-lab/locate/connection"
+	"github.com/m-lab/locate/static"
+)
+
+var locate string
+
+func init() {
+	flag.StringVar(&locate, "locate-url", "ws://localhost:8080/v2/platform/heartbeat/",
+		"URL for locate service")
+}
+
+func main() {
+	flag.Parse()
+	rtx.Must(flagx.ArgsFromEnvWithLog(flag.CommandLine, false), "Failed to read args from env")
+
+	conn := connection.NewConn()
+	conn.Dial(locate, http.Header{})
+	write(conn)
+}
+
+// write starts a write loop to send health messages every
+// HeartbeatPeriod.
+func write(ws *connection.Conn) {
+	defer ws.Close()
+	ticker := time.NewTicker(static.HeartbeatPeriod)
+	defer ticker.Stop()
+
+	for {
+		<-ticker.C
+		err := ws.WriteMessage(websocket.TextMessage, []byte("Health message!"))
+		if err != nil {
+			log.Printf("failed to write, err: %v", err)
+		}
+	}
+}
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -1,0 +1,1 @@
+package main
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -1,1 +1,27 @@
 package main
+
+import (
+	"flag"
+	"testing"
+	"time"
+
+	"github.com/m-lab/locate/connection/testdata"
+)
+
+func Test_main(t *testing.T) {
+	s := testdata.FakeServer()
+	defer s.Close()
+
+	flag.Set("locate-url", s.URL)
+
+	timer := time.NewTimer(time.Second)
+	go func() {
+		<-timer.C
+		if locate != s.URL {
+			t.Error("main() incorrect locate URL")
+		}
+		done <- true
+	}()
+
+	main()
+}
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -1,31 +1,31 @@
 package main
 
-// import (
-// 	"context"
-// 	"flag"
-// 	"testing"
-// 	"time"
+import (
+	"context"
+	"flag"
+	"testing"
+	"time"
 
-// 	"github.com/m-lab/locate/connection/testdata"
-// )
+	"github.com/m-lab/locate/connection/testdata"
+)
 
-// func Test_main(t *testing.T) {
-// 	mainCtx, mainCancel = context.WithCancel(context.Background())
-// 	fh := testdata.FakeHandler{}
-// 	s := testdata.FakeServer(fh.Upgrade)
-// 	defer s.Close()
+func Test_main(t *testing.T) {
+	mainCtx, mainCancel = context.WithCancel(context.Background())
+	fh := testdata.FakeHandler{}
+	s := testdata.FakeServer(fh.Upgrade)
+	defer s.Close()
 
-// 	flag.Set("locate-url", s.URL)
+	flag.Set("locate-url", s.URL)
 
-// 	timer := time.NewTimer(time.Second)
-// 	go func() {
-// 		<-timer.C
-// 		if locate != s.URL {
-// 			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
-// 				locate, s.URL)
-// 		}
-// 		mainCancel()
-// 	}()
+	timer := time.NewTimer(time.Second)
+	go func() {
+		<-timer.C
+		if locate != s.URL {
+			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
+				locate, s.URL)
+		}
+		mainCancel()
+	}()
 
-// 	main()
-// }
+	main()
+}
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -25,7 +25,9 @@ func Test_main(t *testing.T) {
 			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
 				locate, s.URL)
 		}
-		if fh.ReadError() != nil {
+
+		msg, err := fh.Read()
+		if msg == nil || err != nil {
 			t.Errorf("main() failed to write heartbeat message")
 		}
 
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -17,20 +17,13 @@ func Test_main(t *testing.T) {
 
 	flag.Set("locate-url", s.URL)
 
-	heartbeatPeriod = 2 * time.Second
-	timer := time.NewTimer(2 * heartbeatPeriod)
+	timer := time.NewTimer(time.Second)
 	go func() {
 		<-timer.C
 		if locate != s.URL {
 			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
 				locate, s.URL)
 		}
-
-		msg, err := fh.Read()
-		if msg == nil || err != nil {
-			t.Errorf("main() failed to write heartbeat message")
-		}
-
 		mainCancel()
 	}()
 
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -18,7 +18,8 @@ func Test_main(t *testing.T) {
 	go func() {
 		<-timer.C
 		if locate != s.URL {
-			t.Error("main() incorrect locate URL")
+			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
+				locate, s.URL)
 		}
 		done <- true
 	}()
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -15,15 +15,15 @@ func Test_main(t *testing.T) {
 	s := testdata.FakeServer(fh.Upgrade)
 	defer s.Close()
 
-	flag.Set("locate-url", s.URL)
+	flag.Set("heartbeat-url", s.URL)
 
 	heartbeatPeriod = 2 * time.Second
 	timer := time.NewTimer(2 * heartbeatPeriod)
 	go func() {
 		<-timer.C
-		if locate != s.URL {
+		if heartbeatURL != s.URL {
 			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
-				locate, s.URL)
+				heartbeatURL, s.URL)
 		}
 
 		msg, err := fh.Read()
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -1,31 +1,31 @@
 package main
 
-import (
-	"context"
-	"flag"
-	"testing"
-	"time"
+// import (
+// 	"context"
+// 	"flag"
+// 	"testing"
+// 	"time"
 
-	"github.com/m-lab/locate/connection/testdata"
-)
+// 	"github.com/m-lab/locate/connection/testdata"
+// )
 
-func Test_main(t *testing.T) {
-	mainCtx, mainCancel = context.WithCancel(context.Background())
-	fh := testdata.FakeHandler{}
-	s := testdata.FakeServer(fh.Upgrade)
-	defer s.Close()
+// func Test_main(t *testing.T) {
+// 	mainCtx, mainCancel = context.WithCancel(context.Background())
+// 	fh := testdata.FakeHandler{}
+// 	s := testdata.FakeServer(fh.Upgrade)
+// 	defer s.Close()
 
-	flag.Set("locate-url", s.URL)
+// 	flag.Set("locate-url", s.URL)
 
-	timer := time.NewTimer(time.Second)
-	go func() {
-		<-timer.C
-		if locate != s.URL {
-			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
-				locate, s.URL)
-		}
-		mainCancel()
-	}()
+// 	timer := time.NewTimer(time.Second)
+// 	go func() {
+// 		<-timer.C
+// 		if locate != s.URL {
+// 			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
+// 				locate, s.URL)
+// 		}
+// 		mainCancel()
+// 	}()
 
-	main()
-}
+// 	main()
+// }
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"testing"
 	"time"
@@ -9,19 +10,26 @@ import (
 )
 
 func Test_main(t *testing.T) {
-	s := testdata.FakeServer()
+	mainCtx, mainCancel = context.WithCancel(context.Background())
+	fh := testdata.FakeHandler{}
+	s := testdata.FakeServer(fh.Upgrade)
 	defer s.Close()
 
 	flag.Set("locate-url", s.URL)
 
-	timer := time.NewTimer(time.Second)
+	heartbeatPeriod = 2 * time.Second
+	timer := time.NewTimer(2 * heartbeatPeriod)
 	go func() {
 		<-timer.C
 		if locate != s.URL {
 			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
 				locate, s.URL)
 		}
-		done <- true
+		if fh.ReadError() != nil {
+			t.Errorf("main() failed to write heartbeat message")
+		}
+
+		mainCancel()
 	}()
 
 	main()
--- a/cmd/heartbeat/main_test.go
+++ b/cmd/heartbeat/main_test.go
@@ -17,13 +17,20 @@ func Test_main(t *testing.T) {
 
 	flag.Set("locate-url", s.URL)
 
-	timer := time.NewTimer(time.Second)
+	heartbeatPeriod = 2 * time.Second
+	timer := time.NewTimer(2 * heartbeatPeriod)
 	go func() {
 		<-timer.C
 		if locate != s.URL {
 			t.Errorf("main() incorrect locate URL; got: %s, want: %s",
 				locate, s.URL)
 		}
+
+		msg, err := fh.Read()
+		if msg == nil || err != nil {
+			t.Errorf("write() did not send heartbeat message")
+		}
+
 		mainCancel()
 	}()
 
--- a/connection/connection.go
+++ b/connection/connection.go
@@ -4,6 +4,7 @@ package connection
 
 import (
 	"errors"
+	"fmt"
 	"log"
 	"net/http"
 	"net/url"
@@ -16,12 +17,9 @@ import (
 )
 
 var (
-	// ErrNotConnected is returned when the application tries to
-	// read/write a message and the connection is closed.
-	ErrNotConnected = errors.New("websocket not connected")
-	// ErrCannotReconnect is returned when the number of reconnections
+	// ErrTooManyReconnects is returned when the number of reconnections
 	// has reached MaxReconnectionsTotal within MaxElapsedTime.
-	ErrCannotReconnect = errors.New("websocket cannot reconnect right now (too many attemps)")
+	ErrTooManyReconnects = errors.New("websocket cannot reconnect right now (too many attemps)")
 	// ErrNotConnected is returned when WriteMessage is called, but
 	// the websocket has not been created yet.
 	ErrNotDailed = errors.New("websocket not created yet, please call Dial()")
@@ -29,6 +27,17 @@ var (
 	// become successful if the request is retried.
 	retryClientErrors = map[int]bool{404: true, 408: true, 425: true}
 )
+
+// ErrCannotReconnect is returned when the application tries to
+// read/write a message, but the connection is closed and
+// a reconnection attempt has failed.
+type ErrCannotReconnect struct {
+	Reason error
+}
+
+func (e *ErrCannotReconnect) Error() string {
+	return fmt.Sprintf("websocket failed to reconnect, err: %v", e.Reason)
+}
 
 // Conn contains the state needed to connect, reconnect, and send
 // messages.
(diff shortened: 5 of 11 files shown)


cristinaleonr (commented):
I don't know why coveralls is only passing every other build even though the coverage is not changing.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64#-:-N0rDxI30whFZlaknYTM:brqfb18)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`cmd/heartbeat/main_test.go` line 24 at r3](https://reviewable.io/reviews/m-lab/locate/64#-N0mS-vrAxDWyA3D2b91:-N0rAJUJEs0fl36eHwL8:berqo43) ([raw file](https://github.com/m-lab/locate/blob/32c87aa410ad81a662b0de396e9b4ccc466f44de/cmd/heartbeat/main_test.go#L24)):*

I recommend closing the channel. That has the same effect in the select statement, and conveys in code that the channel is not intended to be reused.

In other packages/services we have defined a global `mainCtx` with a cancel function. This would allow the `write` function to accept a context now (which I imagine it will want int he future too as it becomes more elaborate and eventually leaves `main.go` for it's own package.
</blockquote></details>

Thanks, I replaced this with a `mainCtx` with a cancel function. 
___
*[`connection/connection_test.go` line 112 at r3](https://reviewable.io/reviews/m-lab/locate/64#-N0mVkEA1YQOPm8ZoYNI:-N0rDYpa4bSkIfC2r3KP:bcmddoy) ([raw file](https://github.com/m-lab/locate/blob/32c87aa410ad81a662b0de396e9b4ccc466f44de/connection/connection_test.go#L112)):*

What do you think about blackbox testing this package?

What if the FakeServer accepted a "fake handler" (defined in this file) that closed the connection it received?
</blockquote></details>

I like this idea. 

I added a `FakeHandler`. I put it in `fakeserver.go` so I can also use it in `cmd/heartbeat/main_test.go`.

stephen-soltesz (changes requested):
> Reviewed 2 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/main.go` line 24 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N145Tdw3grVMxAu4I0Q:-N145Tdw3grVMxAu4I0R:b-bg93sy) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/cmd/heartbeat/main.go#L24)):*
> ```Go
> 
> func init() {
> 	flag.StringVar(&locate, "locate-url", "ws://localhost:8080/v2/platform/heartbeat/",
> ```

To prevent confusion with ndt or other clients using the "nearest" query with flag named `-locate-url` (or similar), this could be the `-heartbeat-url` flag?
___
*[`cmd/heartbeat/README.md` line 4 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14625-DJhxI4QCNanD:-N14625-DJhxI4QCNanE:b-ld2xxw) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/cmd/heartbeat/README.md#L4)):*
> ```Markdown
> 
> The Heartbeat Service is a sidecar service that performs health checks
> for experiment instances running in the same Kubernetes pod.
> ```

... and reports them to the Locate API. (or similar).
___
*[`cmd/heartbeat/README.md` line 8 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N146ErU4MbmLXg2UHZQ:-N146ErU4MbmLXg2UHZR:ba7vtv2) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/cmd/heartbeat/README.md#L8)):*
> ```Markdown
> On start, it establishes a WebSocket connection with the Locate Service
> and sends an initial registration message. Subsequently, it performs
> periodic instance health checks and sends the results to the Locate.
> ```

*Locate Service or Locate API at the end, yeah?
___
*[`connection/connection.go` line 1 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14Ls2S9lwArQcADCeA:-N14Ls2S9lwArQcADCeB:bwcmj57) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L1)):*
> ```Go
> // Package connection provides a Websocket that will automatically
> ```

I've added several low-level comments. This is a high-level comment, added after all the smaller comments.

This is the most important package in this PR. Here you are defining the interface for a "reconnecting websocket type". Interface design is one of the "arts" of software engineering. I don't claim to have answers to these questions. I am not even sure they're all of the right questions. I'm trying to step back with you to look at the interface together to see what we see.

It looks like the most important parts of that interface are in `Dial` and `WriteMessage`. Consider the following:

* how often do we expect a caller to call `Dial`?
* under what conditions will `Dial` return a fatal error?
* how often will a caller call `WriteMessage`?
* does WriteMessage guarantee delivery? should it? can it?
* how should the caller treat errors returned by `WriteMessage`?
___
*[`connection/connection.go` line 24 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14FLMl8yE6AjrBrWzV:-N14FLMl8yE6AjrBrWzW:b5a4mcn) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L24)):*
> ```Go
> // Conn contains state needed to connect, reconnect, and send
> // messages.
> type Conn struct {
> ```

Which of these fields must be public?

Please add doc string hints about their expected usage if they remain public.
___
*[`connection/connection.go` line 71 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14MO80-W4WEw8fuQSm:-N14MO80-W4WEw8fuQSn:b-mpkdu4) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L71)):*
> ```Go
> 	c.ticker = *time.NewTicker(c.MaxReconnectionsTime)
> 	go func(c *Conn) {
> 		defer c.ticker.Stop()
> ```

This looks like a leak. Under what conditions would this be called?
___
*[`connection/connection.go` line 86 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14KLMW7cYrQzEASYR8:-N14KLMW7cYrQzEASYR9:b-r7vmz4) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L86)):*
> ```Go
> // close and reconnect.
> func (c *Conn) WriteMessage(messageType int, data []byte) error {
> 	if c.IsConnected() {
> ```

Does https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow apply here also? If there was a considered reason to put the positive case first, please add a comment. Otherwise, I can imagine trying to reconnect first since `closeAndReconnect` is blocking, then this function could be a little more likely to succeed.
___
*[`connection/connection.go` line 88 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14IbZ3CUhHSQ-mER4Z:-N14IbZ3CUhHSQ-mER4_:b-9foofw) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L88)):*
> ```Go
> 	if c.IsConnected() {
> 		// It is the call to NextWriter that detects the connection
> 		// has been closed by updating writeErr. Calling WriteMessage
> ```

What does `writeErr` refer to? I don't see that defined or referenced anywhere else.
___
*[`connection/connection.go` line 98 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14Liy56sTpppBKvmVh:-N14Liy56sTpppBKvmVi:b1h9s7p) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L98)):*
> ```Go
> 			c.closeAndReconnect()
> 		}
> 		return err
> ```

Are these errors ever fatal for the caller? Can a caller differentiate these from other package defined errors?
___
*[`connection/connection.go` line 100 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14LMsx-w36jFOSgqtZ:-N14LMsx-w36jFOSgqt_:b-cdmn0u) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L100)):*
> ```Go
> 		return err
> 	}
> 	if c.canConnect() {
> ```

Is the "not connected" and "can't connect" cases worth differentiating in the returned error?
___
*[`connection/connection.go` line 108 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14L0DG4XoEKKQBTW-_:-N14L0DG4XoEKKQBTW-a:b-lh30jq) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L108)):*
> ```Go
> // Close closes the underlying network connection without
> // sending or waiting for a close frame.
> func (c *Conn) Close() {
> ```

Please return any errors returned by `ws.Close`.
___
*[`connection/connection.go` line 165 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14F8P49FWwnQ-4qaaa:-N14F8P49FWwnQ-4qaab:b-1qlu3m) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L165)):*
> ```Go
> 		ws, _, err = c.Dialer.Dial(c.url.String(), c.header)
> 		if err != nil {
> 			log.Printf("could not establish a connection with %s (will retry), err: %v",
> ```

Does this dial loop honor the reconnection limits? Should it? What if I pass a bad url? The `resp` could help diagnose fatal conditions.
___
*[`connection/connection.go` line 167 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14GsLi6GybeF8siG2L:-N14GsLi6GybeF8siG2M:b-9sif7n) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L167)):*
> ```Go
> 			log.Printf("could not establish a connection with %s (will retry), err: %v",
> 				c.url.String(), err)
> 		} else {
> ```

Consider whether it's possible to outdent the "success" case - https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow e.g. by using `continue` within the error case.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`cmd/heartbeat/main.go` line 24 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N145Tdw3grVMxAu4I0Q:-N1d6ZJo7Ci6WD-zv2vu:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/cmd/heartbeat/main.go#L24)):*

To prevent confusion with ndt or other clients using the "nearest" query with flag named `-locate-url` (or similar), this could be the `-heartbeat-url` flag?
</blockquote></details>

Done.
___
*[`cmd/heartbeat/README.md` line 4 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14625-DJhxI4QCNanD:-N1d6_baEFp9_0bOaL-7:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/cmd/heartbeat/README.md#L4)):*

... and reports them to the Locate API. (or similar).
</blockquote></details>

Done.
___
*[`cmd/heartbeat/README.md` line 8 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N146ErU4MbmLXg2UHZQ:-N1d6aTOCUA1pB5Lxedi:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/cmd/heartbeat/README.md#L8)):*

*Locate Service or Locate API at the end, yeah?
</blockquote></details>

Done.
___
*[`connection/connection.go` line 1 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14Ls2S9lwArQcADCeA:-N1eWz_K1yavZXzpsL_A:bcs99ou) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L1)):*
> how often do we expect a caller to call Dial?

Only on start to create the connection. Or, if the caller calls `Close` and wishes to recreate the connection, it will have to call `Dial` again.

> under what conditions will Dial return a fatal error?

The function returns an error if the url is invalid or if a 4XX error (except 404, 408, 425) is received in the HTTP response. These errors are not fatal and the client could try to dial with a different URL/headers.

> how often will a caller call WriteMessage?

In this case, the caller will call `WriteMessage` every 10 secs. Let me know if this is not what you are asking.

> does WriteMessage guarantee delivery? should it? can it?

I wish. It does not and I don't see how it could, though I'm open to suggestions :). 

> how should the caller treat errors returned by WriteMessage?

My intention now with this function is for the majority of the retrying to be done by `WriteMessage` and for there to be little left for the client to do.
* If the error is `ErrNotDialed`, call `Dial`.
* If the error is `ErrTooManyReconnects`, wait until `CanConnect` returns true before trying again.
* If the error is a client error (e.g., the write failed because the message was invalid), I don't know that there is anything the client could do at runtime other than abandon the request.

I have added more comments in the code to clarify these questions.
___
*[`connection/connection.go` line 24 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14FLMl8yE6AjrBrWzV:-N1d6c-92RsdoLjeiORn:bs3r2jz) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L24)):*

Which of these fields must be public?

Please add doc string hints about their expected usage if they remain public.
</blockquote></details>

Added doc strings for the public fields.
___
*[`connection/connection.go` line 71 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14MO80-W4WEw8fuQSm:-N1d6hElEQiB5ipaKDDN:bldnlyx) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L71)):*

This looks like a leak. Under what conditions would this be called?
</blockquote></details>

Moved this to the `Close()` function.
___
*[`connection/connection.go` line 86 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14KLMW7cYrQzEASYR8:-N1iOhgQ63Q_HFlxijOt:b1y48pv) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L86)):*

Does https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow apply here also? If there was a considered reason to put the positive case first, please add a comment. Otherwise, I can imagine trying to reconnect first since `closeAndReconnect` is blocking, then this function could be a little more likely to succeed.
</blockquote></details>

For simplicity, my intention was for the write to fail if there was a disconnect. The function would reconnect and return a write error, which the client could interpret to try to write again. Disconnects are initially detected from write failures anyway, so it made sense to try the write before the reconnect.

Given some of your comments regarding interpreting errors in the caller, I now believe it's cleaner for `WriteMessage` itself to reconnect and resend the message. So, I have moved the error handling to the beginning now, followed by the write.
___
*[`connection/connection.go` line 88 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14IbZ3CUhHSQ-mER4Z:-N1dAUr676_xMvIFlOya:bi42spi) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L88)):*

What does `writeErr` refer to? I don't see that defined or referenced anywhere else.
</blockquote></details>

I have added the [reference](https://github.com/gorilla/websocket/blob/master/conn.go#L489) to the comment.
___
*[`connection/connection.go` line 98 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14Liy56sTpppBKvmVh:-N1dDoy62bInX-12u_bj:bxwj5h) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L98)):*

Are these errors ever fatal for the caller? Can a caller differentiate these from other package defined errors?
</blockquote></details>

Most of the errors defined in the `gorilla/websocket` package have to do with closed connections. There are two other errors, [`errBadWriteOpcode`](https://github.com/gorilla/websocket/blob/78cf1bc733a927f673fd1988a25256b425552a8a/conn.go#L179) and [`errInvalidControlFrame`](https://github.com/gorilla/websocket/blob/78cf1bc733a927f673fd1988a25256b425552a8a/conn.go#L181), that would be good for the caller to be able to recognize, since they indicate issues with the message itself. However, they are not exported. 

I'm not sure what you mean by fatal but I don't believe they should cause the connection to close.
___
*[`connection/connection.go` line 100 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14LMsx-w36jFOSgqtZ:-N1dC9Mv3U9gJtFwPq_q:b-3c001e) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L100)):*

Is the "not connected" and "can't connect" cases worth differentiating in the returned error?
</blockquote></details>

Added new package errors for the different scenarios.
___
*[`connection/connection.go` line 108 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14L0DG4XoEKKQBTW-_:-N1dCEBuEnC2FTx08T3B:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L108)):*

Please return any errors returned by `ws.Close`.
</blockquote></details>

Done.
___
*[`connection/connection.go` line 165 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14F8P49FWwnQ-4qaaa:-N1dF9EVB8lFwleZEUPF:bnqz3hr) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L165)):*

Does this dial loop honor the reconnection limits? Should it? What if I pass a bad url? The `resp` could help diagnose fatal conditions.
</blockquote></details>

It does not honor the reconnection limits. The intention of the reconnection limits was to prevent scenarios where we managed to reconnect and then the connection was soon dropped (which is different from not being able to connect at all, but could also overload the server). 

We can set up a hard limit on the maximum amount of time this loop will run through the `MaxElapsedTime` variable, but I believe it's better to continue to try to reconnect since the service is pretty much useless unless it can communicate with the server.

If you pass in a bad url, `Dial` will return an error. I added handling for different client errors from the response.
___
*[`connection/connection.go` line 167 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14GsLi6GybeF8siG2L:-N1dO3fk8mLWTNLRyWKv:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L167)):*

Consider whether it's possible to outdent the "success" case - https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow e.g. by using `continue` within the error case.
</blockquote></details>

Done.

stephen-soltesz (changes requested):
Checkpointing my comments for faster feedback. Thank you for differentiating the write errors. I need to look more closely at the unit tests next.

> Reviewed 2 of 6 files at r7, 1 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64#-:-N1jYTsL4SXJk1uaYVqt:b-gg2yzv)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`connection/connection.go` line 98 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14Liy56sTpppBKvmVh:-N1jUuZF0ReitwDTZlIn:b-9755fa) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L98)):*

Most of the errors defined in the `gorilla/websocket` package have to do with closed connections. There are two other errors, [`errBadWriteOpcode`](https://github.com/gorilla/websocket/blob/78cf1bc733a927f673fd1988a25256b425552a8a/conn.go#L179) and [`errInvalidControlFrame`](https://github.com/gorilla/websocket/blob/78cf1bc733a927f673fd1988a25256b425552a8a/conn.go#L181), that would be good for the caller to be able to recognize, since they indicate issues with the message itself. However, they are not exported. 

I'm not sure what you mean by fatal but I don't believe they should cause the connection to close.
</blockquote></details>

Thank you.
___
*[`connection/connection.go` line 165 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14F8P49FWwnQ-4qaaa:-N1jWD2c76VjcsKuJGo6:b63l8yn) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L165)):*

It does not honor the reconnection limits. The intention of the reconnection limits was to prevent scenarios where we managed to reconnect and then the connection was soon dropped (which is different from not being able to connect at all, but could also overload the server). 

We can set up a hard limit on the maximum amount of time this loop will run through the `MaxElapsedTime` variable, but I believe it's better to continue to try to reconnect since the service is pretty much useless unless it can communicate with the server.

If you pass in a bad url, `Dial` will return an error. I added handling for different client errors from the response.
</blockquote></details>

What would happen if we deployed a bad API key, and the API returned 404s from every pod?
___
*[`connection/connection.go` line 34 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jR9pUDUSE5XnO4CvY:-N1jR9pUDUSE5XnO4CvZ:bhxtg7j) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L34)):*
> ```Go
> // read/write a message, but the connection is closed and
> // a reconnection attempt has failed.
> type ErrCannotReconnect struct {
> ```

I remember we discovered that this pattern was necessary for middleware management of error types that needed to be preserved from a lower level to the caller. Is this necessary in this case? If the caller wants to compare error values, two above are static and this is a struct. Could we equally/safely use a static `ErrCannotReconnect` type with `%s %w` combining the low level error string with the wrapped static error here?
___
*[`connection/connection.go` line 81 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jSu5t3b4ARzSx6L6H:-N1jSu5t3b4ARzSx6L6I:bmmnx6b) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L81)):*
> ```Go
> 	mu                   sync.Mutex
> 	reconnections        int
> 	dialed               bool
> ```

`isDialed`?
___
*[`connection/connection.go` line 125 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jTbpWByZafa4Iayt3:-N1jTbpWByZafa4Iayt4:b-2wggd1) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L125)):*
> ```Go
> 	c.ticker = *time.NewTicker(c.MaxReconnectionsTime)
> 	go func(c *Conn) {
> 		for {
> ```

Should this loop end after `Conn.Close()`?
___
*[`connection/connection.go` line 277 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jXJ1SBfFiVZORvl73:-N1jXJ1SBfFiVZORvl74:b-v08vb3) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L277)):*
> ```Go
> 	w, err := c.ws.NextWriter(messageType)
> 	if err == nil {
> 		err = c.ws.WriteMessage(messageType, data)
> ```

Should this use the writer `w`?
___
*[`connection/connection_test.go` line 15 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jXg_X-fDWvU1e0y3V:-N1jXg_X-fDWvU1e0y3W:b-sytrcp) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection_test.go#L15)):*
> ```Go
> )
> 
> func Test_Dial(t *testing.T) {
> ```

Please add a test with an explicit Dial, Close, Dial, Close sequence on the same connection.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`connection/connection.go` line 165 at r6](https://reviewable.io/reviews/m-lab/locate/64#-N14F8P49FWwnQ-4qaaa:-N1ktc7-EmtNNjy0FWmx:bdx1605) ([raw file](https://github.com/m-lab/locate/blob/08a55341013d0939436f7ef7b92d5a0d88245b87/connection/connection.go#L165)):*

What would happen if we deployed a bad API key, and the API returned 404s from every pod?
</blockquote></details>

I just tried making a request with an incorrect API key (e.g., https://locate.measurementlab.net/v2/priority/nearest/ndt/ndt7?api_key=AIzaSyAzYTrIbsiOqf294LWJLJ7NDBL91W_yxi) and got a 400 error, in which case we would not continue trying to connect and return the error to the caller (also see the `Test_Dial_BadRequest` unit test).
___
*[`connection/connection.go` line 34 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jR9pUDUSE5XnO4CvY:-N1l3ItF0vK87yJByXGs:bf2hkxy) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L34)):*

I remember we discovered that this pattern was necessary for middleware management of error types that needed to be preserved from a lower level to the caller. Is this necessary in this case? If the caller wants to compare error values, two above are static and this is a struct. Could we equally/safely use a static `ErrCannotReconnect` type with `%s %w` combining the low level error string with the wrapped static error here?
</blockquote></details>

I thought it might be nice to return a custom/identifiable error to the caller when the problem was specific to reconnecting. Now that you mention it though, I wouldn't say it's necessary, so I have removed it and only returned the low level error. 

Your suggestion is good but I think it's more helpful to return the low level error (e.g., `ErrTooManyReconnects`), as opposed to just the string.
___
*[`connection/connection.go` line 81 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jSu5t3b4ARzSx6L6H:-N1ktRJf7xMsSpXhva8H:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L81)):*

`isDialed`?
</blockquote></details>

Done.
___
*[`connection/connection.go` line 125 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jTbpWByZafa4Iayt3:-N1ktUoy6KN5yKsqdfEM:bu7lsvx) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L125)):*

Should this loop end after `Conn.Close()`?
</blockquote></details>

Of course! Added a channel to close this.
___
*[`connection/connection.go` line 277 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jXJ1SBfFiVZORvl73:-N1kvnEY8X5lWYQcrO_w:b-b7h5a1) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L277)):*

Should this use the writer `w`?
</blockquote></details>

Nope, I tried to make my comment more clear. 

The call to `NextWriter` is just to update the internal write error (which then causes the `WriteMessage` call to fail). If the connection is dropped here, neither the `NextWriter` call itself nor a call to `Write` with that same writer `w` will throw an error.

I'm happy to show you an example over vc.
___
*[`connection/connection_test.go` line 15 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jXg_X-fDWvU1e0y3V:-N1ktpv0Af0sgQCkUqWw:bqmjwcz) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection_test.go#L15)):*

Please add a test with an explicit Dial, Close, Dial, Close sequence on the same connection.
</blockquote></details>

Added.

stephen-soltesz (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`connection/connection.go` line 277 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jXJ1SBfFiVZORvl73:-N1oOy5Y4Jfszbb_Vo0c:b-s4sy1u) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L277)):*

Nope, I tried to make my comment more clear. 

The call to `NextWriter` is just to update the internal write error (which then causes the `WriteMessage` call to fail). If the connection is dropped here, neither the `NextWriter` call itself nor a call to `Write` with that same writer `w` will throw an error.

I'm happy to show you an example over vc.
</blockquote></details>

I still don't understand. Why do we need `NextWriter`? `NextWriter` returns an error generated by an earlier call to WriteMessage faster than WriteMessage alone?
___
*[`connection/connection.go` line 31 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oWUmPA3Ji1Yc8Eloz:-N1oWUmPA3Ji1Yc8Elp-:b-newa2s) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection.go#L31)):*
> ```Go
> 
> // Conn contains the state needed to connect, reconnect, and send
> // messages.
> ```

Perhaps emphasize that default values must be updated before calling `Dial`.
___
*[`connection/connection.go` line 92 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oU2pp6foJ_09fvA-m:-N1oU2pp6foJ_09fvA-n:biwjfw3) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection.go#L92)):*
> ```Go
> //
> // The function returns an error if the url is invalid or if
> // a 4XX error (except 404, 408, 425) is received in the HTTP
> ```

Why do we want to retry on 404s? One of my other questions was based on me thinking this was "Not Authorized" but it's "Not Found".
___
*[`connection/connection.go` line 168 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oQn_17dPJCZ_zAAie:-N1oQn_17dPJCZ_zAAif:brfvqc0) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection.go#L168)):*
> ```Go
> }
> 
> // Close closes the "stop" channel and the underlying
> ```

I don't know where it's written - this isn't my idea - including implementation details in the docstring for a public function is almost always an anti-pattern (always exceptions). The caller doesn't need to know about the 'stop' channel since this is not a public field. The caller just needs to know that this function must be called to close the network connection and clean up all private resources (the goroutine) after the connection is done. This rule doesn't apply to private methods.
___
*[`connection/connection_test.go` line 32 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oXNfXAmNmLo92DCHB:-N1oXNfY7bRtt30GR0HV:b-9wksn5) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L32)):*
> ```Go
> }
> 
> func Test_Dial_ThenClose(t *testing.T) {
> ```

Thank you - this was to emphasize that the close wants to end all goroutines/other resources created by Dial. :+1:
___
*[`connection/connection_test.go` line 91 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oS8WyEM_FXYTlXRht:-N1oS8WyEM_FXYTlXRhu:bbmmmz0) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L91)):*
> ```Go
> 			err := c.Dial(tt.url, http.Header{})
> 
> 			if err == nil {
> ```

Is the `https` case not a valid URL?

I think this test is combining two things: url parsing and connection errors. Please make these cases more obviously different. Or, the `Test_Dial_ServerDown` tests does the second part more safely.. nit: port 46311 is not guaranteed to be not-open. 
___
*[`connection/connection_test.go` line 131 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oVOtY-xupEhuqcKAG:-N1oVOtY-xupEhuqcKAH:b-qavljv) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L131)):*
> ```Go
> 		name       string
> 		disconnect bool
> 		wantErr    bool
> ```

wantErr is not used in this test.
___
*[`connection/connection_test.go` line 178 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oVest3tgFoxDr0L5P:-N1oVest3tgFoxDr0L5Q:bxr9yf2) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L178)):*
> ```Go
> }
> 
> func Test_WriteMessage_ErrTooManyReconnects(t *testing.T) {
> ```

Cool :+1:
___
*[`connection/connection_test.go` line 243 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oWjPo-DQjCw53oR8h:-N1oWjPo-DQjCw53oR8i:bmg9ewb) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L243)):*
> ```Go
> 	// is reset.
> 	timer := time.NewTimer(2 * c.MaxReconnectionsTime)
> 	<-timer.C
> ```

just clarifying - this will be a 4sec delay for the unit test?

stephen-soltesz (commented):
My last round of comments are increasingly refinements/clarifications. Still looking good :+1:

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64#-:-N1oYNTA73M0g9CntM-v:b-5owfha)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr and @stephen-soltesz)

___
*[`connection/connection.go` line 277 at r8](https://reviewable.io/reviews/m-lab/locate/64#-N1jXJ1SBfFiVZORvl73:-N1p10gq1bLEloZQj3as:b5tivo4) ([raw file](https://github.com/m-lab/locate/blob/3bc8504dadb7cfd72456836cdfd2edff11958281/connection/connection.go#L277)):*

I still don't understand. Why do we need `NextWriter`? `NextWriter` returns an error generated by an earlier call to WriteMessage faster than WriteMessage alone?
</blockquote></details>

I demoed the behavior with and without `NextWriter` over vc. I have updated the comment to clarify the motivation behind the `NextWriter` call.
___
*[`connection/connection.go` line 31 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oWUmPA3Ji1Yc8Eloz:-N1odWWi41B0Y-xX9b8n:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection.go#L31)):*

Perhaps emphasize that default values must be updated before calling `Dial`.
</blockquote></details>

Done.
___
*[`connection/connection.go` line 92 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oU2pp6foJ_09fvA-m:-N1odeAM35FOQhiJgxuq:b-kt596g) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection.go#L92)):*

Why do we want to retry on 404s? One of my other questions was based on me thinking this was "Not Authorized" but it's "Not Found".
</blockquote></details>

I went through the list of [client errors](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors) and picked the ones the seemed like they might succeed at some point in the future. 

I have removed it, so we won't retry for 404.
___
*[`connection/connection.go` line 168 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oQn_17dPJCZ_zAAie:-N1odTXj7xXVdfxK_RBE:b-bw8e14) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection.go#L168)):*

I don't know where it's written - this isn't my idea - including implementation details in the docstring for a public function is almost always an anti-pattern (always exceptions). The caller doesn't need to know about the 'stop' channel since this is not a public field. The caller just needs to know that this function must be called to close the network connection and clean up all private resources (the goroutine) after the connection is done. This rule doesn't apply to private methods.
</blockquote></details>

Updated.
___
*[`connection/connection_test.go` line 91 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oS8WyEM_FXYTlXRht:-N1ob8k86NY95tOUx7SD:bts28oe) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L91)):*

Is the `https` case not a valid URL?

I think this test is combining two things: url parsing and connection errors. Please make these cases more obviously different. Or, the `Test_Dial_ServerDown` tests does the second part more safely.. nit: port 46311 is not guaranteed to be not-open. 
</blockquote></details>

The `https` case is not a valid URL. Only the `ws` and `wss` schemes are allowed. 

Both cases are testing url parsing, so it returns an error before trying to connect.

I changed the port to 1234.
___
*[`connection/connection_test.go` line 131 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oVOtY-xupEhuqcKAG:-N1o_xLjDOC_yervvJ15:bgcp0wf) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L131)):*

wantErr is not used in this test.
</blockquote></details>

Right, no error is expected here anymore. Removed.
___
*[`connection/connection_test.go` line 243 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oWjPo-DQjCw53oR8h:-N1o_ZC_9d4da-FPaVYY:bejgfc2) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L243)):*

just clarifying - this will be a 4sec delay for the unit test?
</blockquote></details>

Yes, I have decreased it to two seconds now.

stephen-soltesz (changes requested):
My remaining concern is with retrying 5xx errors.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64#-:-N1pDtE53yCGKu6k5fiA:b-ke9kz5)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`connection/connection.go` line 234 at r10](https://reviewable.io/reviews/m-lab/locate/64#-N1pCJaBAQHP-sreSPQ4:-N1pCJaBAQHP-sreSPQ5:bb4txya) ([raw file](https://github.com/m-lab/locate/blob/8fb0d7b13e06047639ddee556218b260d07fd81b/connection/connection.go#L234)):*
> ```Go
> 			if resp != nil {
> 				_, retry := retryClientErrors[resp.StatusCode]
> 				if resp.StatusCode >= 400 && resp.StatusCode <= 451 && !retry {
> ```

To simplify this condition, is the StatusCode range check necessary? For almost all error status codes, `!retry` will be true.

I don't believe we should unconditionally retry all 5xx errors.

Also, https://pkg.go.dev/net/http#Client.Do "A non-2xx status code doesn't cause an error." -- does this apply to websocket dialer?
___
*[`connection/connection_test.go` line 91 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oS8WyEM_FXYTlXRht:-N1pAdFfB_AU4e37KnKz:b-o01jwq) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L91)):*

The `https` case is not a valid URL. Only the `ws` and `wss` schemes are allowed. 

Both cases are testing url parsing, so it returns an error before trying to connect.

I changed the port to 1234.
</blockquote></details>

Thank you for updating the test name also. 
___
*[`connection/connection_test.go` line 131 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oVOtY-xupEhuqcKAG:-N1pAmFv1M7WelXIhQFM:b-j4nyle) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L131)):*

Right, no error is expected here anymore. Removed.
</blockquote></details>

Please also remove `wantErr` from the tests struct.

cristinaleonr (commented):
Thank you for all the review :pray: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64#-:-N1tvhQyEIgF7d75dqCT:b-23ft8t)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`connection/connection.go` line 234 at r10](https://reviewable.io/reviews/m-lab/locate/64#-N1pCJaBAQHP-sreSPQ4:-N1tpvtx8faVQ_9nfsh6:bvx5s0t) ([raw file](https://github.com/m-lab/locate/blob/8fb0d7b13e06047639ddee556218b260d07fd81b/connection/connection.go#L234)):*

To simplify this condition, is the StatusCode range check necessary? For almost all error status codes, `!retry` will be true.

I don't believe we should unconditionally retry all 5xx errors.

Also, https://pkg.go.dev/net/http#Client.Do "A non-2xx status code doesn't cause an error." -- does this apply to websocket dialer?
</blockquote></details>

I have removed the range check and added specific 5XX errors (500, 502, 503, and 504) to `retryErrors`.

According to the documentation for [`websocket.DialContext`](https://pkg.go.dev/github.com/gorilla/websocket#Dialer.DialContext), "If the WebSocket handshake fails, ErrBadHandshake is returned along with a non-nil *http.Response so that callers can handle redirects, authentication, etcetera."
___
*[`connection/connection_test.go` line 131 at r9](https://reviewable.io/reviews/m-lab/locate/64#-N1oVOtY-xupEhuqcKAG:-N1tqp0GElLFC5G-SWO-:b-39v3cp) ([raw file](https://github.com/m-lab/locate/blob/57335787aec20a469d2520f43b9747fd450d0ef6/connection/connection_test.go#L131)):*

Please also remove `wantErr` from the tests struct.
</blockquote></details>

Whoops, fixed.

stephen-soltesz (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/64#-:-N1ty6X583e1syt2hylT:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

PR #68, opened 2022-06-09
Load and send registration message

cristinaleonr:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/68)


stephen-soltesz (commented):
(sorry meant to press publish earlier)

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68#-:-N4EoAxQ16GIYq6Ydz4d:b4i6ud)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/registration/registration.go` line 60 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EMfD3AN-4SDP-vLg0:-N4EMfD3AN-4SDP-vLg1:b67z9j0) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/cmd/heartbeat/registration/registration.go#L60)):*
> ```Go
> 
> 	for _, r := range registrations {
> 		if r.Site == h.Site && r.Machine == h.Machine {
> ```

Ah, recently @nkinkade added a per-node format for the uuid-annotator and (since there is one record per node) defined a map object so that the load and lookup is a simple `map[key]` read. Would that be worth while here?
___
*[`connection/connection.go` line 233 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EVjNH3JXcCZ0aGoR5:-N4EVjNH3JXcCZ0aGoR6:bfztpoj) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/connection/connection.go#L233)):*
> ```Go
> 	for range ticker.C {
> 		if !c.isConnected {
> 			ws, resp, err = c.dialer.Dial(c.url.String(), c.header)
> ```

This block nests four levels deep. Could this be a helper function or simplified?
___
*[`connection/connection.go` line 241 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EVWql2C4781gNXONa:-N4EVWql2C4781gNXONb:bmwm8no) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/connection/connection.go#L241)):*
> ```Go
> 						log.Printf("error trying to establish a connection with %s, err: %v, status: %d",
> 							c.url.String(), err, resp.StatusCode)
> 						return err
> ```

Is there a resource leak on the ticker for this error path?
___
*[`handler/heartbeat.go` line 21 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EU9vN55DfZIF4eb9e:-N4EU9vN55DfZIF4eb9f:b927hmx) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/handler/heartbeat.go#L21)):*
> ```Go
> type HeartbeatMessage struct {
> 	MsgType string           `json:"msgType"`
> 	Msg     *json.RawMessage `json:"msg"`
> ```

I've established today that you have good reasons for the things you've done. Can you say how you arrived at this message format?

I expected to find a single struct, with an embedded `RegistrationMessage` type that was `Marshal`d once before sending by the client and then `Unmarshal`'d once when received by the handler, where the marshal happens as late as possible in the client (just before sending) and unmarshal as early as possible (just after receive).

I'll try to find a good example / reference for what I'm thinking of.
___
*[`handler/heartbeat.go` line 92 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EU0Ay6vq4DuBlwG7i:-N4EU0Ay6vq4DuBlwG7j:bdhmw2z) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/handler/heartbeat.go#L92)):*
> ```Go
> 				}
> 				instance.health = hm.Score
> 			case "default":
> ```

`default:` ?

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68)* status: 0 of 1 approvals obtained (waiting on @nkinkade and @stephen-soltesz)

___
*[`connection/connection.go` line 233 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EVjNH3JXcCZ0aGoR5:-N4XSNpv-eGcT6CKonUt:b5b9lm7) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/connection/connection.go#L233)):*

This block nests four levels deep. Could this be a helper function or simplified?
</blockquote></details>

Simplified. 
___
*[`connection/connection.go` line 241 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EVWql2C4781gNXONa:-N4XSgJh2hOSu4CsnZ5c:b-k2rwtz) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/connection/connection.go#L241)):*

Is there a resource leak on the ticker for this error path?
</blockquote></details>

Good catch. Fixed.
___
*[`handler/heartbeat.go` line 21 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EU9vN55DfZIF4eb9e:-N4XTSlFCA6ZV3232gk5:br6xk8p) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/handler/heartbeat.go#L21)):*

I've established today that you have good reasons for the things you've done. Can you say how you arrived at this message format?

I expected to find a single struct, with an embedded `RegistrationMessage` type that was `Marshal`d once before sending by the client and then `Unmarshal`'d once when received by the handler, where the marshal happens as late as possible in the client (just before sending) and unmarshal as early as possible (just after receive).

I'll try to find a good example / reference for what I'm thinking of.
</blockquote></details>

The reason I was using a message containing a `msgType` and a`json.RawMessage` was that there were two types of messages I needed to differentiate between (e.g., health and registration), and `json.RawMessage` allowed me to decode the message type without having to decode the rest of the message. [Here](https://stackoverflow.com/questions/23053563/how-can-you-decode-multiple-message-types-with-golang-websockets) is a similar example.

Since we only have two message types, I have decided to simplify the message and use a `registered` variable to keep track of which type of message to decode.
___
*[`handler/heartbeat.go` line 92 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EU0Ay6vq4DuBlwG7i:-N4XSpyzA04qB0z_x93i:bh1v0vu) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/handler/heartbeat.go#L92)):*

`default:` ?
</blockquote></details>

Whoops, yes. This code has been removed anyway.
___
*[`cmd/heartbeat/registration/registration.go` line 60 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EMfD3AN-4SDP-vLg0:-N4XSGj11sUIme3Mj161:b2vbwsj) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/cmd/heartbeat/registration/registration.go#L60)):*

Ah, recently @nkinkade added a per-node format for the uuid-annotator and (since there is one record per node) defined a map object so that the load and lookup is a simple `map[key]` read. Would that be worth while here?
</blockquote></details>

Sure, I've udpated siteinfo https://github.com/m-lab/siteinfo/pull/240.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`connection/connection.go` line 238 at r5](https://reviewable.io/reviews/m-lab/locate/68#-N4XfMqPDyUHFQeWoCrG:-N4Y3GbI2t8DXtxdZSff:bfbhcep) ([raw file](https://github.com/m-lab/locate/blob/729a340e09e2064c80e0e5e37841cc27bf9213a6/connection/connection.go#L238)):*

Does the for condition break? iirc, ticker `Stop` does not close the channel, and I would expect the for loop to block, since there is no return here now.
</blockquote></details>

Yes, it does exit. This is what the `Test_Dial_BadRequest` test is testing for in `connection_test.go`. 

Furthermore, since the return has also been removed from the "success" path, all the connection tests would hang/fail.
___
*[`handler/heartbeat.go` line 21 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EU9vN55DfZIF4eb9e:-N4Y52Cu7QrKJgdRuLZc:bulzyyj) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/handler/heartbeat.go#L21)):*

Consider a structure that includes two fields, of specific types. When the JSON is unmarshaled the field will be either complete or nil, indicating the message type.

e.g.

```
type HeartbeatMessage struct {
   Health       *messages.Health
   Registration *messages.Registration
}
```

This would permit potentially adding future messages without additional state variables.
</blockquote></details>

I think the current flow is helpful to ensure that the registration message is the very first one that is received.

It's possible I don't recall discussions of future messages. Did you have any in mind? Otherwise, I think [YAGNI](https://betterprogramming.pub/yagni-you-aint-gonna-need-it-f9a178cd8e1).
___
*[`cmd/heartbeat/messaging/heartbeatmsg.go` line 15 at r5](https://reviewable.io/reviews/m-lab/locate/68#-N4Xbq0Z4p_y4I1BNKnO:-N4Y0MAo40s7-mrsMjfv:b-ksqym5) ([raw file](https://github.com/m-lab/locate/blob/729a340e09e2064c80e0e5e37841cc27bf9213a6/cmd/heartbeat/messaging/heartbeatmsg.go#L15)):*

I want to say this structure is part of the locate API. The same structure needs to be agreed on by both client and server.

Please place this structure under https://github.com/m-lab/locate/tree/master/api/v2

The remaining logic to load registration messages makes sense under the cmd/heartbeat, b/c that's the only command expected to use this.
</blockquote></details>

I've moved the loading logic to cmd/heartbeat and moved the message structs to api/v2 (can't keep the `Health` struct in the `main` package because then it can't be referenced by other packages).

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`connection/connection.go` line 238 at r5](https://reviewable.io/reviews/m-lab/locate/68#-N4XfMqPDyUHFQeWoCrG:-N4Z-bGs7tu5C5-Kp117:burne4c) ([raw file](https://github.com/m-lab/locate/blob/729a340e09e2064c80e0e5e37841cc27bf9213a6/connection/connection.go#L238)):*

Yes, it does exit. This is what the `Test_Dial_BadRequest` test is testing for in `connection_test.go`. 

Furthermore, since the return has also been removed from the "success" path, all the connection tests would hang/fail.
</blockquote></details>

Also, looking at the ticker's [documentation](https://pkg.go.dev/github.com/cenkalti/backoff/v4#NewTicker) (i.e.,  "The channel is closed when Stop method is called") and the [implementation](https://github.com/cenkalti/backoff/blob/v4.1.3/ticker.go#L55) (i.e., `t.stopOnce.Do(func() { close(t.stop) })`), they do seem to verify the "breaking" behavior.

stephen-soltesz (commented):
Note: my feedback is increasingly about interface design and what I believe will be code health for the larger system. If this is stalling needed progress, we can find a middle ground to continue iterating.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68#-:-N4cAHTsAl6G6IRhZsBa:b-5ipoaa)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/main.go` line 52 at r6](https://reviewable.io/reviews/m-lab/locate/68#-N4c8lHw-0g1UEQKQ93Q:-N4c8lHw-0g1UEQKQ93R:bmxgf1x) ([raw file](https://github.com/m-lab/locate/blob/2f5defdedc470faa333995fb91126dacfc796d14/cmd/heartbeat/main.go#L52)):*
> ```Go
> 
> 	// Establish a connection.
> 	conn := connection.NewConn(b)
> ```

`connection.Conn` used to be a generic reliable connection type, now it is mixing the semantics of that generic type and the heartbeat protocol.

I'm also struck by the caller needing to be responsible for marshaling the registration object.

Can you imagine a `HeartbeatConn` that includes a connect.Conn, accepts a registration object, encapsulates the registration marshal and provides a similar encapsulated interface for the health message?
___
*[`connection/connection.go` line 238 at r5](https://reviewable.io/reviews/m-lab/locate/68#-N4XfMqPDyUHFQeWoCrG:-N4YlRPw5-8KMeGYLpcU:b9e4kb3) ([raw file](https://github.com/m-lab/locate/blob/729a340e09e2064c80e0e5e37841cc27bf9213a6/connection/connection.go#L238)):*

Also, looking at the ticker's [documentation](https://pkg.go.dev/github.com/cenkalti/backoff/v4#NewTicker) (i.e.,  "The channel is closed when Stop method is called") and the [implementation](https://github.com/cenkalti/backoff/blob/v4.1.3/ticker.go#L55) (i.e., `t.stopOnce.Do(func() { close(t.stop) })`), they do seem to verify the "breaking" behavior.
</blockquote></details>

Oh, I see - I'm applying my knowledge of `time.Ticker` to `backoff.Ticker` which has changed the ticker API by closing the channel.

We went to lengths to ensure that the memoryless.Ticker was a drop in replacement https://pkg.go.dev/github.com/m-lab/go@v0.1.48/memoryless#NewTicker for the time.Ticker. `backoff` did not do that (for better or worse).
___
*[`handler/heartbeat.go` line 21 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EU9vN55DfZIF4eb9e:-N4c5qwY-Dqsil040jNC:b-e49k08) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/handler/heartbeat.go#L21)):*

I think the current flow is helpful to ensure that the registration message is the very first one that is received.

It's possible I don't recall discussions of future messages. Did you have any in mind? Otherwise, I think [YAGNI](https://betterprogramming.pub/yagni-you-aint-gonna-need-it-f9a178cd8e1).
</blockquote></details>

Okay - the struct may not be the key issue. Something doesn't feel right yet.
___
*[`handler/heartbeat.go` line 43 at r6](https://reviewable.io/reviews/m-lab/locate/68#-N4c9ZSs3fDgQL9dTS40:-N4c9ZSs3fDgQL9dTS41:b-1qjhla) ([raw file](https://github.com/m-lab/locate/blob/2f5defdedc470faa333995fb91126dacfc796d14/handler/heartbeat.go#L43)):*
> ```Go
> 
> // read handles incoming messages from the connection.
> func read(ws *websocket.Conn) {
> ```

The `read` method depends on two global values and a local value on the stack..

Can you imagine an encapsulating object created by Heartbeat handler that manages the order of registration & health updates as today, but references instance fields instead of globals?

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`cmd/heartbeat/main.go` line 52 at r6](https://reviewable.io/reviews/m-lab/locate/68#-N4c8lHw-0g1UEQKQ93Q:-N4epK1U7SOYNPyHPzbO:bupjje4) ([raw file](https://github.com/m-lab/locate/blob/2f5defdedc470faa333995fb91126dacfc796d14/cmd/heartbeat/main.go#L52)):*
> connection.Conn used to be a generic reliable connection type, now it is mixing the semantics of that generic type and the heartbeat protocol.

I totally agree with keeping the connection generic. While `registration` might have been a poor naming choice, my intention is for the message to be an initialization message that any type of caller can specify. I got the thought from [this comment thread](https://docs.google.com/document/d/1SsiycY1qfL8ln87mCU43FG0zduV6bK5Dd4F7TN4r7ps/edit?resourcekey=0-EsrvuketAPB65LdftSDoFw&disco=AAAAT6utlRs).

I have changed its name to `dialMessage` and moved it to the `Dial` function since its purpose is more accurately captured in the context of opening up the connection.

> I'm also struck by the caller needing to be responsible for marshaling the registration object.

I don't see the problem with this, especially in light of the goal of keeping the connection generic. 

The "[gorilla/websocke](https://pkg.go.dev/github.com/gorilla/websocket#Conn.WriteMessage)t" package takes in a `[]byte`, as does the [first reconnecting library](https://pkg.go.dev/github.com/Bitcopy/recws#RecConn.WriteMessage) I considered using. 

>Can you imagine a HeartbeatConn that includes a connect.Conn, accepts a registration object, encapsulates the registration marshal and provides a similar encapsulated interface for the health message?

I implemented this, and then I deleted it in favor of taking in a generic message of type `interface{}` as a parameter, which is more flexible. Messages of this type are automatically encoded by the underlying `WriteJSON` function in the "gorilla/websocket" package.
___
*[`handler/heartbeat.go` line 21 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EU9vN55DfZIF4eb9e:-N4ewsDC3_s8rNcY1-HC:b43dl59) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/handler/heartbeat.go#L21)):*

Okay - the struct may not be the key issue. Something doesn't feel right yet.
</blockquote></details>

Feel free to elaborate.
___
*[`handler/heartbeat.go` line 43 at r6](https://reviewable.io/reviews/m-lab/locate/68#-N4c9ZSs3fDgQL9dTS40:-N4ewyzF3tGT3VIJfLS3:b-2ivcd) ([raw file](https://github.com/m-lab/locate/blob/2f5defdedc470faa333995fb91126dacfc796d14/handler/heartbeat.go#L43)):*

The `read` method depends on two global values and a local value on the stack..

Can you imagine an encapsulating object created by Heartbeat handler that manages the order of registration & health updates as today, but references instance fields instead of globals?
</blockquote></details>

I added the global values as fields in the `Client` struct.

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/main.go` line 52 at r6](https://reviewable.io/reviews/m-lab/locate/68#-N4c8lHw-0g1UEQKQ93Q:-N4ggKv93YDDivwrgkRu:b-kvnkuv) ([raw file](https://github.com/m-lab/locate/blob/2f5defdedc470faa333995fb91126dacfc796d14/cmd/heartbeat/main.go#L52)):*

> connection.Conn used to be a generic reliable connection type, now it is mixing the semantics of that generic type and the heartbeat protocol.

I totally agree with keeping the connection generic. While `registration` might have been a poor naming choice, my intention is for the message to be an initialization message that any type of caller can specify. I got the thought from [this comment thread](https://docs.google.com/document/d/1SsiycY1qfL8ln87mCU43FG0zduV6bK5Dd4F7TN4r7ps/edit?resourcekey=0-EsrvuketAPB65LdftSDoFw&disco=AAAAT6utlRs).

I have changed its name to `dialMessage` and moved it to the `Dial` function since its purpose is more accurately captured in the context of opening up the connection.

> I'm also struck by the caller needing to be responsible for marshaling the registration object.

I don't see the problem with this, especially in light of the goal of keeping the connection generic. 

The "[gorilla/websocke](https://pkg.go.dev/github.com/gorilla/websocket#Conn.WriteMessage)t" package takes in a `[]byte`, as does the [first reconnecting library](https://pkg.go.dev/github.com/Bitcopy/recws#RecConn.WriteMessage) I considered using. 

>Can you imagine a HeartbeatConn that includes a connect.Conn, accepts a registration object, encapsulates the registration marshal and provides a similar encapsulated interface for the health message?

I implemented this, and then I deleted it in favor of taking in a generic message of type `interface{}` as a parameter, which is more flexible. Messages of this type are automatically encoded by the underlying `WriteJSON` function in the "gorilla/websocket" package.

</blockquote></details>

Thank you. The problem I saw with the caller marshaling the registration object was an abstraction boundary violation. The client would need to know about details that are too low-level. Using `WriteJSON` is a great way to put the marshal later. Now the client only needs to know about the `Registration` type rather than the wire format.
___
*[`handler/heartbeat.go` line 21 at r2](https://reviewable.io/reviews/m-lab/locate/68#-N4EU9vN55DfZIF4eb9e:-N4gbjg26NKGwqQbRPc2:b-1pxti9) ([raw file](https://github.com/m-lab/locate/blob/3883ff87e5dd0891d647e1ab4049e6646c869a27/handler/heartbeat.go#L21)):*

Feel free to elaborate.
</blockquote></details>

We spoke in person.
___
*[`handler/heartbeat.go` line 54 at r12](https://reviewable.io/reviews/m-lab/locate/68#-N4ge9vm6JwEUXH3Of6K:-N4ge9vm6JwEUXH3Of6L:bph66qt) ([raw file](https://github.com/m-lab/locate/blob/fa95d3eaad1ce665269a9784e1f9a2ebbb920550/handler/heartbeat.go#L54)):*
> ```Go
> 			if !registered {
> 				var rm v2.Registration
> 				if err := json.Unmarshal(message, &rm); err != nil {
> ```

The current logic ensures that the handler attempts to decode a registration message first. How does it ensure that `message` actually contains a registration message?

Please add a test case that sends a wrong message type that the handler handles accordingly.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`handler/heartbeat.go` line 54 at r12](https://reviewable.io/reviews/m-lab/locate/68#-N4ge9vm6JwEUXH3Of6K:-N4jmZ0P4_ts83LFHLTl:b-azek0g) ([raw file](https://github.com/m-lab/locate/blob/fa95d3eaad1ce665269a9784e1f9a2ebbb920550/handler/heartbeat.go#L54)):*

The current logic ensures that the handler attempts to decode a registration message first. How does it ensure that `message` actually contains a registration message?

Please add a test case that sends a wrong message type that the handler handles accordingly.
</blockquote></details>

I added the `HeartbeatMessage` struct with pointers to the message types so the program can easily check which field is not `nil`.

I also added a test with an invalid message type (`TestClient_Heartbeat_InvalidRegistrationType`).

stephen-soltesz (commented):
One thought, one suggestion for simplifying the public API by removing the `Health.Hostname` field.

This PR is shaping up well. Thank you.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68#-:-N4mYLVkEJkrDisO6cGO:b-qcrq0o)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`handler/handler.go` line 38 at r13](https://reviewable.io/reviews/m-lab/locate/68#-N4mXl8o-Fisak4BI0uY:-N4mXl8o-Fisak4BI0uZ:b-4t683j) ([raw file](https://github.com/m-lab/locate/blob/073716b1dca6b457ead0fe87933875776f483e32/handler/handler.go#L38)):*
> ```Go
> 	ClientLocator
> 	targetTmpl *template.Template
> 	instances  map[string]*instanceData
> ```

FYI: Begin thinking about the interface that the Heartbeat handler and ultimately from the Nearest handler will use to access this information.
___
*[`handler/heartbeat.go` line 38 at r13](https://reviewable.io/reviews/m-lab/locate/68#-N4mUX0p0fuYiq5BonYV:-N4mUX0p0fuYiq5BonYW:byextye) ([raw file](https://github.com/m-lab/locate/blob/073716b1dca6b457ead0fe87933875776f483e32/handler/heartbeat.go#L38)):*
> ```Go
> 
> // handleHeartbeats handles incoming messages from the connection.
> func (c *Client) handleHeartbeats(ws *websocket.Conn) {
> ```

Does the Health message need to include the `Hostname` field? The `handleHearbeats` goroutine will always be for a single host source. Without the `Health.Hostname` the `instanceData` struct begins to look very similar to `v2.HeartbeatMessage`. Could `v2.HeartbeatMessage` be used for internal accounting also?
___
*[`handler/heartbeat.go` line 77 at r13](https://reviewable.io/reviews/m-lab/locate/68#-N4mVo5s4ybwi9onarkd:-N4mVo5s4ybwi9onarke:btncf0j) ([raw file](https://github.com/m-lab/locate/blob/073716b1dca6b457ead0fe87933875776f483e32/handler/heartbeat.go#L77)):*
> ```Go
> 	defer c.mu.Unlock()
> 	if instance, found := c.instances[hm.Hostname]; found {
> 		instance.health = hm.Score
> ```

Please use the same field name for the same value/information/data. #notranslations

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`handler/heartbeat.go` line 38 at r13](https://reviewable.io/reviews/m-lab/locate/68#-N4mUX0p0fuYiq5BonYV:-N4mmdb86McXVmwvRTBY:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/073716b1dca6b457ead0fe87933875776f483e32/handler/heartbeat.go#L38)):*

Does the Health message need to include the `Hostname` field? The `handleHearbeats` goroutine will always be for a single host source. Without the `Health.Hostname` the `instanceData` struct begins to look very similar to `v2.HeartbeatMessage`. Could `v2.HeartbeatMessage` be used for internal accounting also?
</blockquote></details>

Done.
___
*[`handler/heartbeat.go` line 77 at r13](https://reviewable.io/reviews/m-lab/locate/68#-N4mVo5s4ybwi9onarkd:-N4mmekg5txKQORJkTmv:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/073716b1dca6b457ead0fe87933875776f483e32/handler/heartbeat.go#L77)):*

Please use the same field name for the same value/information/data. #notranslations
</blockquote></details>

Done.

stephen-soltesz (commented):
Some code health comments in heartbeat_test.go

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68#-:-N4mqgpN0KMDYKMk_VGe:bv9x1oa)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`handler/heartbeat_test.go` line 29 at r15](https://reviewable.io/reviews/m-lab/locate/68#-N4mpbay9BDMbGlgL0-j:-N4mpbay9BDMbGlgL0-k:bjp6giv) ([raw file](https://github.com/m-lab/locate/blob/3ac7e98c3892cb7c9f6f0907a8f32c398a75af10/handler/heartbeat_test.go#L29)):*
> ```Go
> 	u.Scheme = "ws"
> 	dialer := websocket.Dialer{}
> 	ws, _, err := dialer.Dial(u.String(), http.Header{})
> ```

Is `ws` guaranteed to be safe to call `ws.Close()` on if `err != nil`?

This could be a `mustSetupTest` function (if it helps).
___
*[`handler/heartbeat_test.go` line 81 at r15](https://reviewable.io/reviews/m-lab/locate/68#-N4mp5Cq5qZoNAydTzf_:-N4mp5Cq5qZoNAydTzfa:b-tbljl7) ([raw file](https://github.com/m-lab/locate/blob/3ac7e98c3892cb7c9f6f0907a8f32c398a75af10/handler/heartbeat_test.go#L81)):*
> ```Go
> 	c.mu.RLock()
> 	defer c.mu.RUnlock()
> 	val, _ := c.instances[testdata.FakeHostname]
> ```

`, _` can be omitted. Go supports both forms of map index:

```
val := c.instances[key]
// or 
val, ok := c.instances[key]
```
___
*[`handler/heartbeat_test.go` line 99 at r15](https://reviewable.io/reviews/m-lab/locate/68#-N4moLsO1yvqzeMWoRwm:-N4moLsO1yvqzeMWoRwn:bvxrwm) ([raw file](https://github.com/m-lab/locate/blob/3ac7e98c3892cb7c9f6f0907a8f32c398a75af10/handler/heartbeat_test.go#L99)):*
> ```Go
> 	r := testdata.FakeRegistration
> 	r.Registration.Latitude = math.Inf(1)
> 	ws.WriteJSON(r)
> ```

I expect this to fail on marshal not unmarshal (as implied by function name).

Please check `WriteJSON` error values throughout. We have https://pkg.go.dev/github.com/m-lab/go@v0.1.48/testingx#Must if it's helpful.

Is there a way to verify the server is receiving & rejecting the message?

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`handler/heartbeat_test.go` line 29 at r15](https://reviewable.io/reviews/m-lab/locate/68#-N4mpbay9BDMbGlgL0-j:-N4nH6K85doI6NgGmfUt:b-hkc9uc) ([raw file](https://github.com/m-lab/locate/blob/3ac7e98c3892cb7c9f6f0907a8f32c398a75af10/handler/heartbeat_test.go#L29)):*

Is `ws` guaranteed to be safe to call `ws.Close()` on if `err != nil`?

This could be a `mustSetupTest` function (if it helps).
</blockquote></details>

Added an error check.
___
*[`handler/heartbeat_test.go` line 81 at r15](https://reviewable.io/reviews/m-lab/locate/68#-N4mp5Cq5qZoNAydTzf_:-N4nHOoCAkrpUL-xGck8:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/3ac7e98c3892cb7c9f6f0907a8f32c398a75af10/handler/heartbeat_test.go#L81)):*

`, _` can be omitted. Go supports both forms of map index:

```
val := c.instances[key]
// or 
val, ok := c.instances[key]
```
</blockquote></details>

Done.
___
*[`handler/heartbeat_test.go` line 99 at r15](https://reviewable.io/reviews/m-lab/locate/68#-N4moLsO1yvqzeMWoRwm:-N4nHSPW3i6g2-i_CnYT:bs4vafl) ([raw file](https://github.com/m-lab/locate/blob/3ac7e98c3892cb7c9f6f0907a8f32c398a75af10/handler/heartbeat_test.go#L99)):*

I expect this to fail on marshal not unmarshal (as implied by function name).

Please check `WriteJSON` error values throughout. We have https://pkg.go.dev/github.com/m-lab/go@v0.1.48/testingx#Must if it's helpful.

Is there a way to verify the server is receiving & rejecting the message?
</blockquote></details>

You're right. These are the remnants of an old test that I tried to keep when I switched to `WriteJSON`.

I have added error checks for the `WriteJSON` calls.

There is a test to verify that valid messages are received and handled appropriately and a test to verify that invalid messages don't have unwanted side effects. I'm happy to look into further verification in the future.

stephen-soltesz (approved):
:lgtm: 🎉

> Reviewed 8 of 12 files at r6, 2 of 4 files at r8, 1 of 1 files at r9, 1 of 1 files at r10, 5 of 6 files at r14, 1 of 1 files at r16.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/68#-:-N4nNW1VETaCwEfZH4EU:bkt6kjg)* status: :shipit: complete! 1 of 1 approvals obtained

PR #71, opened 2022-06-30
Implement Redis Client 

cristinaleonr:


stephen-soltesz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/71)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`instances/redis.go` line 53 at r2](https://reviewable.io/reviews/m-lab/locate/71#-N6Tulcb5pOAjBAu7Bxd:-N6Tulcb5pOAjBAu7Bxe:bhoqz0a) ([raw file](https://github.com/m-lab/locate/blob/7e6f132f1142f6c0dd8157bf2c7b5d4bd2d1d612/instances/redis.go#L53)):*
> ```Go
> 		return errKeyNotFound
> 	}
> 	return rc.Put(key, field, value)
> ```

iiuc, we're using the `redis.Pool` for concurrent access. It looks like the `conn` for `Update` is open during the call to `Put`, which will open a new `conn`. Depending on how the redis Pool implements that, this could deadlock, no? Is this guaranteed to return multiple conns?

Or, what happens if someone uses a Pool with `MaxActive = 1`?

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/71)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`api/v2/api.go` line 139 at r2](https://reviewable.io/reviews/m-lab/locate/71#-N5qlWJrCERMXxvGNwdb:-N6j3yhb23BcwLbXNam_:bapqmk5) ([raw file](https://github.com/m-lab/locate/blob/7e6f132f1142f6c0dd8157bf2c7b5d4bd2d1d612/api/v2/api.go#L139)):*

Are these tags needed? The default behavior from the json marshaller is to use the field name. So, `Health` should use the json field named `Health` without tags. Does that change when used with redis? Does redis happen to have this default behavior also? In the logic below I only see `json` operations -- iiuc it looks like redis is operating on bytes (e.g. the RedisScan functions). When are the 'redis' tags used?

So, if the default behavior will work, please remove the redundant tags. If they are required for some reason, please add a comment explaining the need.

Maybe only these top level fields so `getHeartbeat` can WAI?
</blockquote></details>

They are not "needed," but missing tags have inadvertently caused inconsistencies in the past, so I thought it might be good to add them here now that these fields will start to be persisted.



I have removed them now. Will add them back if there is a need.
___
*[`instances/redis.go` line 53 at r2](https://reviewable.io/reviews/m-lab/locate/71#-N6Tulcb5pOAjBAu7Bxd:-N6j5F9YDVLWpLY7Wd-w:b-sz16ul) ([raw file](https://github.com/m-lab/locate/blob/7e6f132f1142f6c0dd8157bf2c7b5d4bd2d1d612/instances/redis.go#L53)):*

iiuc, we're using the `redis.Pool` for concurrent access. It looks like the `conn` for `Update` is open during the call to `Put`, which will open a new `conn`. Depending on how the redis Pool implements that, this could deadlock, no? Is this guaranteed to return multiple conns?

Or, what happens if someone uses a Pool with `MaxActive = 1`?
</blockquote></details>

I have removed `Update`, so now only one `conn` is open.
___
*[`instances/redis.go` line 77 at r2](https://reviewable.io/reviews/m-lab/locate/71#-N6TVxKWBRHnj2LT1iqr:-N6j5f7S55hoDK2LGgTz:b7ha9cm) ([raw file](https://github.com/m-lab/locate/blob/7e6f132f1142f6c0dd8157bf2c7b5d4bd2d1d612/instances/redis.go#L77)):*

Do I understand correctly, that an individual ScanStruct operation could fail, but we want to continue for all those that succeed? i.e. one err is not fatal, but the set of results returned by `GetAllHeartbeats` may not be the complete set?

How can callers decide whether a final `err != nil` is fatal or whether the set of results are sufficient? This code embodies a policy that is not consistent with the current docstring. "Partial success" is harder to reason about but may be needed if Redis behaves this way often enough. "All or nothing" approach is simpler to reason about for callers and implementation. Is this an option here?
</blockquote></details>

We discussed this in person. I changed this to an "all or nothing" approach.

PR #80, opened 2022-08-17
Add Kubernetes health checks to the HBS

cristinaleonr:
This PR extends the HBS to contact the Kubernetes API server to check the pod and node health.

It uses the CA certificate and token in the pod's file system to authenticate the requests against the public API.

Changes were tested by pushing a new image to `k8s-support` in sandbox and checking that the `v2beta2` query returns the right servers (and excludes any that are in maintenance).

Note: a lot of the line count comes from `go.mod/go.sum` changes.

This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/80)


stephen-soltesz (commented):
This is great. I'm glad the public-api approach worked out, even if we can't use in-cluster auth yet. Thank you for figuring this out.

I added some first round questions about optionality of the k8s api and how we can monitor the efficiency of these queries in production.

> Reviewed 4 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/80#-:-NAGRNza-vihOIaWkwEZ:bxkib20)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/main.go` line 66 at r2](https://reviewable.io/reviews/m-lab/locate/80#-NAGMJXO0m8NYmWQRDlo:-NAGMJXO0m8NYmWQRDlp:b-72f2sw) ([raw file](https://github.com/m-lab/locate/blob/132d5e0bb08bbd134d3daff1cb1d678c78acd1b0/cmd/heartbeat/main.go#L66)):*
> ```Go
> 
> 	probe := health.NewPortProbe(s)
> 	k8s := health.MustNewKubernetesClient(mainCtx, kubernetesURL.URL, pod, node, namespace, kubernetesAuth)
> ```

Does this require deployment within k8s? Would it be straightforward to make this optional if kubernetes.URL is not specified? e.g. local development or a private deployment outside of a k8s environment?
___
*[`cmd/heartbeat/health/kubernetes-client.go` line 38 at r2](https://reviewable.io/reviews/m-lab/locate/80#-NAGOoMe8hLVms-Y03TN:-NAGOoMe8hLVms-Y03TO:bd4nqii) ([raw file](https://github.com/m-lab/locate/blob/132d5e0bb08bbd134d3daff1cb1d678c78acd1b0/cmd/heartbeat/health/kubernetes-client.go#L38)):*
> ```Go
> 		node:      node,
> 		namespace: namespace,
> 		ctx:       ctx,
> ```

My instinct is that not storing the context is preferable here. [There may be reasons to do it](https://go.dev/blog/context-and-structs) but the [context package documentation](https://pkg.go.dev/context) recommends "Do not store Contexts inside a struct type".

I recommend only passing ctx to functions that use the context at runtime.
___
*[`cmd/heartbeat/health/kubernetes-client.go` line 95 at r2](https://reviewable.io/reviews/m-lab/locate/80#-NAGQXM_-7TfW7j6LTuA:-NAGQXM_-7TfW7j6LTuB:blolvrf) ([raw file](https://github.com/m-lab/locate/blob/132d5e0bb08bbd134d3daff1cb1d678c78acd1b0/cmd/heartbeat/health/kubernetes-client.go#L95)):*
> ```Go
> 
> func (c *KubernetesClient) isPodRunning() bool {
> 	pod, err := c.clientset.CoreV1().Pods(c.namespace).Get(c.ctx, c.pod, metav1.GetOptions{})
> ```

These will run every 10 sec?

I'm worried about latency from some locations.

I'm also wondering what will happen if the API server becomes transiently unavailable (maybe during an upgrade).

Please add prometheus histogram metrics for call latency on these requests. Possibly only one around `isHealthy`, differentiating between true/false times. We will want to be able to reason about how frequently these requests take longer than 10sec.
___
*[`cmd/heartbeat/health/kubernetes-client.go` line 118 at r2](https://reviewable.io/reviews/m-lab/locate/80#-NAGR8NN1POeMHeYo2Lr:-NAGR8NN1POeMHeYo2Ls:b-ei65dd) ([raw file](https://github.com/m-lab/locate/blob/132d5e0bb08bbd134d3daff1cb1d678c78acd1b0/cmd/heartbeat/health/kubernetes-client.go#L118)):*
> ```Go
> 
> func (c *KubernetesClient) isInMaintenance() bool {
> 	node, err := c.clientset.CoreV1().Nodes().Get(c.ctx, c.node, metav1.GetOptions{})
> ```

Is is possible to preserve the `node` result from above so only one node API request is needed?

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/80)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`cmd/heartbeat/main.go` line 66 at r2](https://reviewable.io/reviews/m-lab/locate/80#-NAGMJXO0m8NYmWQRDlo:-NAQ-r9n1SvhRu802HFQ:b-o9cevy) ([raw file](https://github.com/m-lab/locate/blob/132d5e0bb08bbd134d3daff1cb1d678c78acd1b0/cmd/heartbeat/main.go#L66)):*

Does this require deployment within k8s? Would it be straightforward to make this optional if kubernetes.URL is not specified? e.g. local development or a private deployment outside of a k8s environment?
</blockquote></details>

Ah, this is a very good point. I have made the k8s checks optional.
___
*[`cmd/heartbeat/health/kubernetes-client.go` line 38 at r2](https://reviewable.io/reviews/m-lab/locate/80#-NAGOoMe8hLVms-Y03TN:-NAQ099KEOfhE18RDpfp:b9d12q5) ([raw file](https://github.com/m-lab/locate/blob/132d5e0bb08bbd134d3daff1cb1d678c78acd1b0/cmd/heartbeat/health/kubernetes-client.go#L38)):*

My instinct is that not storing the context is preferable here. [There may be reasons to do it](https://go.dev/blog/context-and-structs) but the [context package documentation](https://pkg.go.dev/context) recommends "Do not store Contexts inside a struct type".

I recommend only passing ctx to functions that use the context at runtime.
</blockquote></details>

Thanks, I struggled with this question when I was first implementing it. I'm passing it in at runtime now.
___
*[`cmd/heartbeat/health/kubernetes-client.go` line 95 at r2](https://reviewable.io/reviews/m-lab/locate/80#-NAGQXM_-7TfW7j6LTuA:-NAQ0PZq51wjtWD7NTMN:b55qm8l) ([raw file](https://github.com/m-lab/locate/blob/132d5e0bb08bbd134d3daff1cb1d678c78acd1b0/cmd/heartbeat/health/kubernetes-client.go#L95)):*

These will run every 10 sec?

I'm worried about latency from some locations.

I'm also wondering what will happen if the API server becomes transiently unavailable (maybe during an upgrade).

Please add prometheus histogram metrics for call latency on these requests. Possibly only one around `isHealthy`, differentiating between true/false times. We will want to be able to reason about how frequently these requests take longer than 10sec.
</blockquote></details>

Yes, these calls are made every 10 secs.

I added a histogram metric. At least in sandbox, it seems like the calls take <1 sec.


___
*[`cmd/heartbeat/health/kubernetes-client.go` line 118 at r2](https://reviewable.io/reviews/m-lab/locate/80#-NAGR8NN1POeMHeYo2Lr:-NAQ1QhX-8y4r3RE_gf_:bcvjgzp) ([raw file](https://github.com/m-lab/locate/blob/132d5e0bb08bbd134d3daff1cb1d678c78acd1b0/cmd/heartbeat/health/kubernetes-client.go#L118)):*

Is is possible to preserve the `node` result from above so only one node API request is needed?
</blockquote></details>

Good idea. Done.

stephen-soltesz (commented):
This looks great. I saw the most recent updates to handling errors from the k8s api and new metrics. A few more small thoughts.

> Reviewed 1 of 10 files at r1, 3 of 8 files at r3, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/80#-:-NAR0gti3YUWjOjtd19r:b-bo0gno)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/main.go` line 94 at r3](https://reviewable.io/reviews/m-lab/locate/80#-NAR-HftBVxgECsJVH0H:-NAR-HftBVxgECsJVH0I:bce54m9) ([raw file](https://github.com/m-lab/locate/blob/8ad5f7dbb3cb23abaf4895f0b7dfde4c650988c1/cmd/heartbeat/main.go#L94)):*
> ```Go
> 			return
> 		case <-ticker.C:
> 			score := hc.GetHealth(mainCtx)
> ```

The health check could accept a `context.WithTimeout`. With the changes to how errors from the k8s api are handled, this could be safe and guarantee regular updates.
___
*[`cmd/heartbeat/health/checker.go` line 29 at r3](https://reviewable.io/reviews/m-lab/locate/80#-NAQxCfv-PzOHwciW5n0:-NAQxCfv-PzOHwciW5n1:bna5f88) ([raw file](https://github.com/m-lab/locate/blob/8ad5f7dbb3cb23abaf4895f0b7dfde4c650988c1/cmd/heartbeat/health/checker.go#L29)):*
> ```Go
> func (hc *Checker) GetHealth(ctx context.Context) float64 {
> 	if hc.pp.checkPorts() {
> 		if hc.k8s != nil && !hc.k8s.isHealthy(ctx) {
> ```

nit: this is a little hard to follow. Suggest reordering:

```
if !hc.pp.checkPorts() {
   return 0
}
if hc.k8s != nil && !hc.k8s.isHealthy(ctx) {
  return 0
}
return 1
```
___
*[`cmd/heartbeat/health/kubernetes-client.go` line 104 at r5](https://reviewable.io/reviews/m-lab/locate/80#-NAR0NA04RT7jQ1sChik:-NAR0NA04RT7jQ1sChil:b-xe0lf) ([raw file](https://github.com/m-lab/locate/blob/8471166f5d9f53c46f202eef70e9e32ffcfb2f8b/cmd/heartbeat/health/kubernetes-client.go#L104)):*
> ```Go
> 	pod, err := c.clientset.CoreV1().Pods(c.namespace).Get(ctx, c.pod, metav1.GetOptions{})
> 	if err != nil {
> 		metrics.KubernetesRequestsTotal.WithLabelValues(err.Error()).Inc()
> ```

Thank you for this.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/80)* status: 0 of 1 approvals obtained (waiting on @stephen-soltesz)

___
*[`cmd/heartbeat/main.go` line 94 at r3](https://reviewable.io/reviews/m-lab/locate/80#-NAR-HftBVxgECsJVH0H:-NARNiod0Zq5lzulwUF6:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/8ad5f7dbb3cb23abaf4895f0b7dfde4c650988c1/cmd/heartbeat/main.go#L94)):*

The health check could accept a `context.WithTimeout`. With the changes to how errors from the k8s api are handled, this could be safe and guarantee regular updates.
</blockquote></details>

Done.
___
*[`cmd/heartbeat/health/checker.go` line 29 at r3](https://reviewable.io/reviews/m-lab/locate/80#-NAQxCfv-PzOHwciW5n0:-NARNwBk6O78LVGNBqGc:bo3vl6l) ([raw file](https://github.com/m-lab/locate/blob/8ad5f7dbb3cb23abaf4895f0b7dfde4c650988c1/cmd/heartbeat/health/checker.go#L29)):*

nit: this is a little hard to follow. Suggest reordering:

```
if !hc.pp.checkPorts() {
   return 0
}
if hc.k8s != nil && !hc.k8s.isHealthy(ctx) {
  return 0
}
return 1
```
</blockquote></details>

Thanks, done!

stephen-soltesz (approved):
:lgtm:

> Reviewed 3 of 5 files at r4, 1 of 2 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/80#-:-NARVWkG7HsK6epuYQPB:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained

___
*[`cmd/heartbeat/main.go` line 107 at r7](https://reviewable.io/reviews/m-lab/locate/80#-NARR7OK18CNGlURjCy_:-NARR7OK18CNGlURjCya:b-bviipm) ([raw file](https://github.com/m-lab/locate/blob/c2d8b325f6d3014947ed80217fd37b0783fb80ad/cmd/heartbeat/main.go#L107)):*
> ```Go
> 
> func getHealth(hc *health.Checker) float64 {
> 	ctx, cancel := context.WithTimeout(mainCtx, heartbeatPeriod)
> ```

Ah, a helper function like this is  a good solution to using `defer`.

PR #82, opened 2022-09-27
Fix dash URL in the static configuration

robertodauria:
neubot/dash currently still uses mlab-ns, but is migrating to locate. @bassosimone expressed a preference toward wrapping dash's endpoint with the access controller directly (rather than using the envelope sidecar mechanism) similarly to what ndt-server does.

DASH's test endpoints  (`/dash/download/<number_of_bytes>`) won't work unless the client has called `/negotiate/dash` first and has received a valid session ID, so securing `/negotiate/dash` makes sure that tests cannot be run without a valid access token.

This PR changes the static config for `neubot/dash` to provide the correct URL (`https://<server>/negotiate/dash?access_token=...`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/82)

### DIFF
--- a/static/configs.go
+++ b/static/configs.go
@@ -55,7 +55,7 @@ var Configs = map[string]Ports{
 		URL("wss", ":3010", "/ndt_protocol"),
 	},
 	"neubot/dash": {
-		URL("https", "", "/"),
+		URL("https", "", "/negotiate/dash"),
 	},
 	"wehe/replay": {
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/82#-:-NCzHRJ525N0mOGB45G5:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained

PR #83, opened 2022-09-27
Add /health endpoint check

cristinaleonr:


nkinkade (changes requested):
> Reviewed 2 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/83)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/health/checker.go` line 37 at r2](https://reviewable.io/reviews/m-lab/locate/83#-ND06awT9reZYqSPbw8e:-ND06awT9reZYqSPbw8f:b-whzq4k) ([raw file](https://github.com/m-lab/locate/blob/13f8decf387d6445b0fec844d66d4d674bb20583/cmd/heartbeat/health/checker.go#L37)):*
> ```Go
> 	}
> 	status, err := checkHealthEndpoint()
> 	if err == nil && !status {
> ```

Just to be sure I understand, here we don't care about the case where `err != nil`, in which case we disregard the health check?
___
*[`cmd/heartbeat/health/checker_test.go` line 135 at r2](https://reviewable.io/reviews/m-lab/locate/83#-ND09yRiC6cD-JGrOWEF:-ND09yRiC6cD-JGrOWEG:b-gdb9d8) ([raw file](https://github.com/m-lab/locate/blob/13f8decf387d6445b0fec844d66d4d674bb20583/cmd/heartbeat/health/checker_test.go#L135)):*
> ```Go
> 		t.Run(tt.name, func(t *testing.T) {
> 			if tt.endpointStatus != 0 {
> 				srv := testHealthServer(tt.endpointStatus)
> ```

Something feels awkward about calling a function in this file (`testHealthServer()`) which is defined in file _endpoint-check_test.go_. Is there some standard way to create testing functions that can be used in multiple test files? I feel like I did this recently by creating a new package named something like `healthtest` and including shared functions and utilities there, but I can't find it now. It seems like [ndt7 might do this](https://github.com/m-lab/ndt-server/tree/main/ndt7/ndt7test) too.
___
*[`cmd/heartbeat/health/endpoint-check.go` line 19 at r2](https://reviewable.io/reviews/m-lab/locate/83#-ND00_aR4v2ZgbFyo38t:-ND00_aR4v2ZgbFyo38u:b-l8ugkm) ([raw file](https://github.com/m-lab/locate/blob/13f8decf387d6445b0fec844d66d4d674bb20583/cmd/heartbeat/health/endpoint-check.go#L19)):*
> ```Go
> 	resp, err := http.Get(healthAddress)
> 	if err != nil {
> 		metrics.HealthEndpointChecksTotal.WithLabelValues(err.Error()).Inc()
> ```

Is this going to put a string describing the error into  to the `status` label of the metric? If so, something feels wrong about that. As long as the set of possible errors and related descriptions is relatively small, would it make sense to add another label for this?
___
*[`cmd/heartbeat/health/endpoint-check.go` line 24 at r2](https://reviewable.io/reviews/m-lab/locate/83#-ND0-u_k1JE1-YeB6CZX:-ND0-u_k1JE1-YeB6CZY:b-xkym2) ([raw file](https://github.com/m-lab/locate/blob/13f8decf387d6445b0fec844d66d4d674bb20583/cmd/heartbeat/health/endpoint-check.go#L24)):*
> ```Go
> 
> 	if resp.StatusCode == http.StatusOK {
> 		metrics.HealthEndpointChecksTotal.WithLabelValues("OK").Inc()
> ```

Would it make more sense to return the status code itself here (e.g., "200") instead of string describing the status code? Same for below in the case that the status code is not 200. Status codes are very well defined and do not change, but the text used to describe them may differ depending on the environment.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/83)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`cmd/heartbeat/health/checker.go` line 37 at r2](https://reviewable.io/reviews/m-lab/locate/83#-ND06awT9reZYqSPbw8e:-ND0d7gg2WD1vailTxob:b-u9b93n) ([raw file](https://github.com/m-lab/locate/blob/13f8decf387d6445b0fec844d66d4d674bb20583/cmd/heartbeat/health/checker.go#L37)):*

Just to be sure I understand, here we don't care about the case where `err != nil`, in which case we disregard the health check?
</blockquote></details>

Correct, because some experiments might not support this check. I have added a comment.
___
*[`cmd/heartbeat/health/checker_test.go` line 135 at r2](https://reviewable.io/reviews/m-lab/locate/83#-ND09yRiC6cD-JGrOWEF:-ND0dDbhBvoyKpjggxCj:by7njx6) ([raw file](https://github.com/m-lab/locate/blob/13f8decf387d6445b0fec844d66d4d674bb20583/cmd/heartbeat/health/checker_test.go#L135)):*

Something feels awkward about calling a function in this file (`testHealthServer()`) which is defined in file _endpoint-check_test.go_. Is there some standard way to create testing functions that can be used in multiple test files? I feel like I did this recently by creating a new package named something like `healthtest` and including shared functions and utilities there, but I can't find it now. It seems like [ndt7 might do this](https://github.com/m-lab/ndt-server/tree/main/ndt7/ndt7test) too.
</blockquote></details>

Good point. I have added a `healthtest` package.
___
*[`cmd/heartbeat/health/endpoint-check.go` line 19 at r2](https://reviewable.io/reviews/m-lab/locate/83#-ND00_aR4v2ZgbFyo38t:-ND0fp77Auo9c1Lo49h-:bsudzc2) ([raw file](https://github.com/m-lab/locate/blob/13f8decf387d6445b0fec844d66d4d674bb20583/cmd/heartbeat/health/endpoint-check.go#L19)):*

Is this going to put a string describing the error into  to the `status` label of the metric? If so, something feels wrong about that. As long as the set of possible errors and related descriptions is relatively small, would it make sense to add another label for this?
</blockquote></details>

You're right.

Another label might be superfluous for the cases where the error is nil.

The nature of the error is less important than differentiating this scenario from the other two below. I have updated this to use a generic error code.
___
*[`cmd/heartbeat/health/endpoint-check.go` line 24 at r2](https://reviewable.io/reviews/m-lab/locate/83#-ND0-u_k1JE1-YeB6CZX:-ND0eDtU7llDk2mnOTmy:bhv9b62) ([raw file](https://github.com/m-lab/locate/blob/13f8decf387d6445b0fec844d66d4d674bb20583/cmd/heartbeat/health/endpoint-check.go#L24)):*

Would it make more sense to return the status code itself here (e.g., "200") instead of string describing the status code? Same for below in the case that the status code is not 200. Status codes are very well defined and do not change, but the text used to describe them may differ depending on the environment.
</blockquote></details>

While I see your point, [here](https://reviewable.io/reviews/m-lab/locate/76#-N80Ewdh0IvWYjCafu04) is also good argument for using human-readable labels. This is also consistent with the `status` label for other metrics in the same package.

What do you think?

nkinkade (approved):
:lgtm:

> Reviewed 4 of 5 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/83#-:-ND3qN_SARawUXS8K3cH:bnfp4nl)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/health/endpoint-check.go` line 24 at r2](https://reviewable.io/reviews/m-lab/locate/83#-ND0-u_k1JE1-YeB6CZX:-ND3oroa1QKBnjftMx4D:bu2nywa) ([raw file](https://github.com/m-lab/locate/blob/13f8decf387d6445b0fec844d66d4d674bb20583/cmd/heartbeat/health/endpoint-check.go#L24)):*

While I see your point, [here](https://reviewable.io/reviews/m-lab/locate/76#-N80Ewdh0IvWYjCafu04) is also good argument for using human-readable labels. This is also consistent with the `status` label for other metrics in the same package.

What do you think?
</blockquote></details>

That makes sense. I too am often looking up HTTP status codes to remember the meaning. I still think it could be useful to have a label with the actual status code, particularly for aggregation of the metric, but if this current usage is consistent with other metrics in this package, as well as other uses in other code bases, then as is seems good.

PR #84, opened 2022-09-28
Make generated tokens unique

robertodauria:
This PR changes the token generation so that the `jwt.Claims.ID` field (which becomes `jti` in JWT) is populated with a new UUID each time.

This makes each generated token unique, which in turn makes them viable as "session IDs" for measurements that make multiple calls across different endpoints/servers, and potentially allows us to check if tokens are being misused.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/84)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/cenkalti/backoff/v4 v4.1.3
 	github.com/go-test/deep v1.0.8
 	github.com/gomodule/redigo v1.8.8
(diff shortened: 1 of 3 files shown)


cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/84#-:-ND3hNhl30ZGvWo_PGMi:bg7igkv)* status: :shipit: complete! 1 of 1 approvals obtained

PR #118, opened 2023-02-21
Add Ready and Live request handlers

stephen-soltesz:
This change adds two new handlers to the Locate v2 for liveness and readiness checks in the App Engine configuration.

With this change:
* /v2/live - unconditionally returns success. This resource tests the minimal live condition for the server.
* /v2/ready - returns success only if the Memorystore import has occurred successfully recently.

These resources are now referenced in new `readiness_check` and `liveness_check` settings in the App Engine app.yaml configuration.

These changes are expected to address https://github.com/m-lab/locate/issues/114

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/118)

Issue #114, opened 2023-02-13
App Engine deployment of Locate with --promote causes brief outage


stephen-soltesz:
With the --promote flag to the deployment of Locate in App Engine, all the heartbeat services are briefly disconnected and must reset with the Locate server before the set of healthy servers are re-populated again. This takes a surprisingly long time.

<img width="1324" alt="Screen Shot 2023-02-13 at 10 29 08 AM" src="https://user-images.githubusercontent.com/1085316/218500245-6818c82b-e4bd-4f77-bf42-f1194befb39a.png">

Unfortunately, Flexible environment App Engine [does not support automatic migration](https://cloud.google.com/appengine/docs/flexible/migrating-traffic) (as Standard environment does).

However, we should still be able to create a tool that performs a gradual migration using Traffic Splitting. 
* https://cloud.google.com/appengine/docs/admin-api/migrating-splitting-traffic
* https://pkg.go.dev/google.golang.org/api@v0.109.0/appengine/v1#Service

It may also be possible to improve the shutdown / warmup mechanism for Locate to "hand off" from one version to the next more gracefully.

stephen-soltesz:
An incremental split appears to work as intended. From 19:35 to 20:10 a 10% increase every 5min resulted in no visible decrease in test rates or locate connections. The increase in traffic was due to hourly client traffic.

<img width="1135" alt="Screen Shot 2023-02-13 at 3 26 04 PM" src="https://user-images.githubusercontent.com/1085316/218567906-4a3f91d0-ed88-4b8a-9707-19b13228104b.png">

stephen-soltesz:
During the original event on 2023-02-09, the time for all servers to re-register took over 1hr (see image below). We see the same slow update in staging. @cristinaleonr suspects this is due to the heartbeat service's exponential backoff and plans to add additional metrics to the hbs so that we can see both the node and locate metrics.

<img width="967" alt="Screen Shot 2023-02-21 at 3 44 44 PM" src="https://user-images.githubusercontent.com/1085316/220454662-2fd13e60-6967-41b1-91ea-82bdc793d081.png">


Also, notable, during the manually split deployment to production on 13th, we do not see disruptions to the available health server counts.

<img width="970" alt="Screen Shot 2023-02-21 at 3 43 39 PM" src="https://user-images.githubusercontent.com/1085316/220454435-c394e1be-eb89-4ecd-979c-2da80690e022.png">

PR #120, opened 2023-02-23
Remove cap for number of reconnections per hour

cristinaleonr:
This PR removes the 10 reconnections per hour cap. It also reduces the reconnection backoff maximum interval to 5 minutes.

When a new version is promoted/created, the heartbeat instances briefly disconnect from the old version and reconnect to the new one. This can sometimes take a surprisingly long time if an instance hits the maximum number of reconnections attempts (10), at which point it has to wait a whole hour to reconnect, or if the retry backoff increases significantly.

This scenario can be reproduced in sandbox. That said, the problem is much less likely to happen since the introduction of the App Engine [custom request handlers](https://github.com/m-lab/locate/pull/118). 

Screenshot 2023-02-27 12 09 40 PM

This is part of the remediation for https://github.com/m-lab/locate/issues/114.

This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/120)

### DIFF
--- a/cloudbuild/cloudbuild.yaml
+++ b/cloudbuild/cloudbuild.yaml
@@ -6,16 +6,16 @@ options:
   - WORKSPACE_LINK=/go/src/github.com/m-lab/locate
 
 steps:
-# Run unit tests for environment.
-- name: gcr.io/$PROJECT_ID/golang-cbif:1.18
-  env:
-  - GONOPROXY=github.com/m-lab/go/*
-  args:
-  - go version
-  - go get -v -t ./...
-  - go vet ./...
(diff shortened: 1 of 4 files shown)


stephen-soltesz (approved):
:lgtm:

We discussed handling 404's differently but it looks like the retries will continue unconditionally (eventually) for any error so that sgtm.

> Reviewed 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/120#-:-NPJAmRuARw1fyt64qwu:b-2os4c9)* status: :shipit: complete! 1 of 1 approvals obtained

PR #126, opened 2023-03-05
Add metro ranking

cristinaleonr:
This PR adds a new `metro_rank` parameter to the output URLs. Specifically, it adds a `Ranks` field to the output of the `LocatorV2.Nearest()` function containing a mapping of machine hostnames to their metro ranking (based on distance from the client).

It also adds the metro ranking to a `locate_metro_distance_ranking` histogram metric.

The URL parameter can be seen [here](https://locate-dot-mlab-sandbox.appspot.com/v2/nearest/ndt/ndt7) and the metric [here](https://prometheus.mlab-sandbox.measurementlab.net/graph?g0.expr=locate_metro_distance_ranking_bucket&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=12h).

This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/126)

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -57,7 +57,7 @@ type Locator interface {
 // LocatorV2 defines how the Nearest handler requests machines nearest to the
 // client.
 type LocatorV2 interface {
-	Nearest(service string, lat, lon float64, opts *heartbeat.NearestOptions) ([]v2.Target, []url.URL, error)
+	Nearest(service string, lat, lon float64, opts *heartbeat.NearestOptions) (*heartbeat.TargetInfo, error)
 	heartbeat.StatusTracker
 }
 
@@ -69,6 +69,12 @@ type ClientLocator interface {
 // PrometheusClient defines the interface to query Prometheus.
 type PrometheusClient interface {
 	Query(ctx context.Context, query string, ts time.Time, opts ...prom.Option) (model.Value, prom.Warnings, error)
+}
+
+type paramOpts struct {
+	raw     url.Values
+	version string
+	ranks   map[string]int
 }
 
 func init() {
@@ -103,17 +109,25 @@ func NewClientDirect(project string, private Signer, locator Locator, locatorV2 
 	}
 }
 
-func extraParams(raw url.Values, version string) url.Values {
+func extraParams(hostname string, p paramOpts) url.Values {
 	v := url.Values{}
 	// Add client parameters.
-	for key := range raw {
+	for key := range p.raw {
 		if strings.HasPrefix(key, "client_") {
 			// note: we only use the first value.
-			v.Set(key, raw.Get(key))
+			v.Set(key, p.raw.Get(key))
 		}
 	}
+
 	// Add Locate Service version.
-	v.Set("locate_version", version)
+	v.Set("locate_version", p.version)
+
+	// Add metro rank.
+	rank, ok := p.ranks[hostname]
+	if ok {
+		v.Set("metro_rank", strconv.Itoa(rank))
+	}
+
 	return v
 }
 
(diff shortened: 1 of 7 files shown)


stephen-soltesz (commented):
One comment from a quick pass. I need a few more minutes for a second pass.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/126#-:-NPrD6il-22N9nKbCeP7:b4s7nwk)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`heartbeat/location.go` line 200 at r1](https://reviewable.io/reviews/m-lab/locate/126#-NPrBqeB88pUgfpu2WRV:-NPrBqeB88pUgfpu2WRW:b-eitvfv) ([raw file](https://github.com/m-lab/locate/blob/813dea61bb549c904582d984a6e10ffa946ca6f1/heartbeat/location.go#L200)):*
> ```Go
> func rank(sites []site) {
> 	metroRank := 0
> 	metros := make(map[string]int)
> ```

We can guarantee the input is sorted.

Can we do this with constant space?

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/126)* status: 0 of 1 approvals obtained

___
*[`heartbeat/location.go` line 200 at r1](https://reviewable.io/reviews/m-lab/locate/126#-NPrBqeB88pUgfpu2WRV:-NPrUAovAwxJ9YLHno8w:b-mb7lv5) ([raw file](https://github.com/m-lab/locate/blob/813dea61bb549c904582d984a6e10ffa946ca6f1/heartbeat/location.go#L200)):*

We can guarantee the input is sorted.

Can we do this with constant space?
</blockquote></details>

Nevermind. Sort by location does not guarantee sort by metro.

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/126)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`heartbeat/location.go` line 200 at r1](https://reviewable.io/reviews/m-lab/locate/126#-NPrBqeB88pUgfpu2WRV:-NPrXGxZ1zgvmW_AljCu:b-dgqfmy) ([raw file](https://github.com/m-lab/locate/blob/813dea61bb549c904582d984a6e10ffa946ca6f1/heartbeat/location.go#L200)):*

Nevermind. Sort by location does not guarantee sort by metro.
</blockquote></details>

Metro used to be equivalent to distinct lat/lon combinations. The trial in India violated that assumption recently, but other locations have done it in the past too. While I've used the expression "metro rank" to describe this idea, I believe it was base on a mistake, since metro != distinct location.

If the sorting is by location, then this could be a "location rank" instead. That is the behavior I intended with the phrase "metro rank". For example, in India currently, the "metro rank" (or whatever we call it) should all be 0, since they're all in the same lat/lon location. But, counting distinct metro names would give the wrong impression about the spread there.

I'm sorry for the confusion. I suggest we call this "location rank" (or target rank), and since the sorting is by distance, we may be able to do this in constant space after all.

stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/126)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`heartbeat/location.go` line 200 at r1](https://reviewable.io/reviews/m-lab/locate/126#-NPrBqeB88pUgfpu2WRV:-NPs5Nwr5WE15-Ce0sWg:b-ezymwe) ([raw file](https://github.com/m-lab/locate/blob/813dea61bb549c904582d984a6e10ffa946ca6f1/heartbeat/location.go#L200)):*

Metro used to be equivalent to distinct lat/lon combinations. The trial in India violated that assumption recently, but other locations have done it in the past too. While I've used the expression "metro rank" to describe this idea, I believe it was base on a mistake, since metro != distinct location.

If the sorting is by location, then this could be a "location rank" instead. That is the behavior I intended with the phrase "metro rank". For example, in India currently, the "metro rank" (or whatever we call it) should all be 0, since they're all in the same lat/lon location. But, counting distinct metro names would give the wrong impression about the spread there.

I'm sorry for the confusion. I suggest we call this "location rank" (or target rank), and since the sorting is by distance, we may be able to do this in constant space after all.


</blockquote></details>

Thank you for starting this change. We discussed in person, and I'll continue this feature.

PR #169, opened 2023-11-16
Add ability to rate limit clients based on a cron schedule

cristinaleonr:
This PR adds a `-limit` flag to the Locate to limit client requests under cron schedules.

Tested locally with docker-compose using `-limit=foo=* * * * *`.

```
$ curl -A "foo" http://127.0.0.1:8080/v2/nearest/ndt/ndt7
{
  "error": {
    "type": "client",
    "title": "Too many periodic requests. Please contact support@measurementlab.net.",
    "status": 429
  }
```

In sandbox, this flag has been added to the build trigger to limit the latest periodic integration every 15 minutes.
```
$ curl -A "node-fetch/1.0 (+https://github.com/bitinn/node-fetch)" https://locate-dot-mlab-sandbox.appspot.com/v2/nearest/ndt/ndt7?region=US-OH
{
  "error": {
    "type": "client",
    "title": "Too many periodic requests. Please contact support@measurementlab.net.",
    "status": 429
  }
```

This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/169)


stephen-soltesz (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/169)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr and @nkinkade)

___
*[`locate.go` line 56 at r1](https://reviewable.io/reviews/m-lab/locate/169#-NjV3EvI8rtC_xfu7dhT:-NjV3EvI8rtC_xfu7dhU:b7j2hh0) ([raw file](https://github.com/m-lab/locate/blob/f80daadda8af507741d3c1d9b6b2444746ae660f/locate.go#L56)):*
> ```Go
> 		Value:   "secretmanager",
> 	}
> 	lmts flagx.KeyValue
> ```

`flagx.KeyValue` allows multiple `key1=value1,key2=value2` to be provided at once. User agent strings with commas will be split or error incorrectly.
___
*[`locate.go` line 77 at r1](https://reviewable.io/reviews/m-lab/locate/169#-NjV2sEh7qJp1QE5sC-d:-NjV2sEh7qJp1QE5sC-e:bc3g4k7) ([raw file](https://github.com/m-lab/locate/blob/f80daadda8af507741d3c1d9b6b2444746ae660f/locate.go#L77)):*
> ```Go
> 	flag.Var(&maxmind, "maxmind-url", "When -locator-maxmind is true, the tar URL of MaxMind IP database. May be: gs://bucket/file or file:./relativepath/file")
> 	flag.Var(&keySource, "key-source", "Where to load signer and verifier keys")
> 	flag.Var(&lmts, "limits", "Cron schedule limits for user agents (agent=schedule).")
> ```

I suggest `agent-limits` as a name to self-describe.
___
*[`cloudbuild/cloudbuild.yaml` line 32 at r1](https://reviewable.io/reviews/m-lab/locate/169#-NjV33Qm9D1NXKqE47uS:-NjV33Qm9D1NXKqE47uT:bnzmd72) ([raw file](https://github.com/m-lab/locate/blob/f80daadda8af507741d3c1d9b6b2444746ae660f/cloudbuild/cloudbuild.yaml#L32)):*
> ```YAML
>     -e 's/{{PLATFORM_PROJECT}}/$_PLATFORM_PROJECT/'
>     -e 's/{{REDIS_ADDRESS}}/$_REDIS_ADDRESS/'
>     -e 's/{{LIMITS}}/$_LIMITS/'
> ```

User agents can include `/` and `,` characters. I would expect `/` to corrupt or confuse the substitution expression.

stephen-soltesz (commented):
Sorry added a few drive by comments based on quick first impressions. @nkinkade this is your PR.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/169#-:-NjV4tIg04yPu6gnKk6G:b66vbl3)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr and @nkinkade)

nkinkade (changes requested):
@cristinaleonr: I added a few comments, in addition to the few @stephen-soltesz added.

> Reviewed 5 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/169#-:-NjiD4FuAm06mLWnfG89:b-jhxl6j)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cloudbuild/app.yaml.template` line 41 at r1](https://reviewable.io/reviews/m-lab/locate/169#-NjiAdbl8bQCU1Ex_qeP:-NjiAdbl8bQCU1Ex_qeQ:blhcxf7) ([raw file](https://github.com/m-lab/locate/blob/f80daadda8af507741d3c1d9b6b2444746ae660f/cloudbuild/app.yaml.template#L41)):*
> ```
>   PROMETHEUSX_LISTEN_ADDRESS: ':9090' # Must match one of the forwarded_ports above.
>   PROMETHEUS_URL: 'https://prometheus-basicauth.{{PLATFORM_PROJECT}}.measurementlab.net/'
>   LIMITS: {{LIMITS}}
> ```

Similar to the comment above from Stephen, maybe these should be `AGENT_LIMITS`, too?
___
*[`handler/handler_test.go` line 591 at r1](https://reviewable.io/reviews/m-lab/locate/169#-NjiBjtu8ehWJYRzzTgG:-NjiBjtu8ehWJYRzzTgH:b-mlz0mr) ([raw file](https://github.com/m-lab/locate/blob/f80daadda8af507741d3c1d9b6b2444746ae660f/handler/handler_test.go#L591)):*
> ```Go
> 			name: "allowed-user-agent-limited-time",
> 			limits: map[string]*limits.Cron{
> 				"foo": limits.NewCron("* * * * *"), // Every minute of every hour.
> ```

Just noting that `cronexpr` allows for 7 cron definition fields (including seconds and years), and will [automatically populate them](https://github.com/aptible/supercronic/tree/master/cronexpr#other-details) if it only finds 5 or 6 fields defined. It may be better to always explicitly define 7 fields. Indeed for the 15m periodic clients in DFW, we will probably want a schedule definition like the following, and we may want unit tests that match down to the second:

`7,8 0,15,30,45 * * * * *` 
___
*[`limits/cron.go` line 26 at r1](https://reviewable.io/reviews/m-lab/locate/169#-Njhsw2S6uB9z_h8CWcK:-NjQRhbZ1vN4Y1dL3yRh:blz92vu) ([raw file](https://github.com/m-lab/locate/blob/f80daadda8af507741d3c1d9b6b2444746ae660f/limits/cron.go#L26)):*
> ```Go
> 
> // IsLimited returns whether the input time is within a time-limited
> // window [start, end).
> ```

Do I understand this function correctly?

* It goes backward in time 10m so that `c.Next()` can zoom forward to the next timestamp matching `schedule`, which in a perfect world would match `t`?
* Then it adds 10m to `start` to allow a window for possible client retries?

If this reading is correct, 10m backward works well for clients with an interval >=10m, but doesn't work as well for clients with a smaller interval. For clients with, say, a 1m interval (like in your unit test), it will zoom back 10m and `c.Next()` would match `t` - 9m and `end` would be `t` + 1m, which still would evaluate to true but would only leave a 1m window to catch client retries. Would it make sense to only go back in time some number of seconds (5 or 10?) such that `c.Next()` would be more or less guaranteed to return a timestamp that might match `t`, and the 10m window for retries will always be about 10m?

nkinkade (changes requested):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/169)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`limits/cron.go` line 26 at r1](https://reviewable.io/reviews/m-lab/locate/169#-Njhsw2S6uB9z_h8CWcK:-Njmmw8L6dkhCk1HCMhk:b-3q16ko) ([raw file](https://github.com/m-lab/locate/blob/f80daadda8af507741d3c1d9b6b2444746ae660f/limits/cron.go#L26)):*

Do I understand this function correctly?

* It goes backward in time 10m so that `c.Next()` can zoom forward to the next timestamp matching `schedule`, which in a perfect world would match `t`?
* Then it adds 10m to `start` to allow a window for possible client retries?

If this reading is correct, 10m backward works well for clients with an interval >=10m, but doesn't work as well for clients with a smaller interval. For clients with, say, a 1m interval (like in your unit test), it will zoom back 10m and `c.Next()` would match `t` - 9m and `end` would be `t` + 1m, which still would evaluate to true but would only leave a 1m window to catch client retries. Would it make sense to only go back in time some number of seconds (5 or 10?) such that `c.Next()` would be more or less guaranteed to return a timestamp that might match `t`, and the 10m window for retries will always be about 10m?
</blockquote></details>

Chatting with @cristinaleonr  yesterday, it turns out that I failed to correctly understand the function. Namely, I failed to realize/understand that going backward in time 10m is mostly to catch retries, not just to rewind for the sake of `c.Next()`.

@cristinaleonr: I was thinking about this a bit more and started wondering whether we could just encode the retry interval into the cron schedule definition? For example, for the 15m periodic client, could we use a cron definition like `* 0-10,15-25,30-40,45-55 * * * * *`? It _seems_ like cronexpr may support this sort of syntax. Granted, for a client that tests at such short intervals it would basically mean we are blocking that user agent almost entirely, so we'd need to be careful that not too many other well-behaved clients are using that same user agent.

nkinkade (approved):
:lgtm: Thank you!

> Reviewed 2 of 7 files at r2, 1 of 9 files at r3, 10 of 12 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/169#-:-Nl0EgHM4HUbHDy-S6-G:b-ae66e1)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @cristinaleonr and @stephen-soltesz)

___
*[`locate.go` line 75 at r6](https://reviewable.io/reviews/m-lab/locate/169#-Nl0B0pb881GcKaj4T7K:-Nl0B0pc8KL1UTspqzNX:b-wfl246) ([raw file](https://github.com/m-lab/locate/blob/28ef3a23c7095c98f271037505b67125ff7f85fa/locate.go#L75)):*
> ```Go
> 	flag.Var(&maxmind, "maxmind-url", "When -locator-maxmind is true, the tar URL of MaxMind IP database. May be: gs://bucket/file or file:./relativepath/file")
> 	flag.Var(&keySource, "key-source", "Where to load signer and verifier keys")
> 	flag.StringVar(&limitsPath, "limits-path", "limits/config.yaml", "Path to the limits config file")
> ```

Is this default path relative to the root directory? It should probably be an absolute path to be on the safe side.
___
*[`limits/config.yaml` line 3 at r6](https://reviewable.io/reviews/m-lab/locate/169#-Nl0EPej4iwM13qgOr0u:-Nl0EPek0RN6JqNCGDJg:b7ciqni) ([raw file](https://github.com/m-lab/locate/blob/28ef3a23c7095c98f271037505b67125ff7f85fa/limits/config.yaml#L3)):*
> ```YAML
> ---
> - agent: "node-fetch/1.0 (+https://github.com/bitinn/node-fetch)"
>   schedule: "7,8 0,15,30,45 * * * * *"
> ```

Just noting that this should catch the periodic client and any retries for up to one minute, but that we might need increase the duration if we find that the client is set to retry for a good while.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/169)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @nkinkade and @stephen-soltesz)

___
*[`locate.go` line 75 at r6](https://reviewable.io/reviews/m-lab/locate/169#-Nl0B0pb881GcKaj4T7K:-Nl0vXBXBZqEybAf3-7k:b-8up8vx) ([raw file](https://github.com/m-lab/locate/blob/28ef3a23c7095c98f271037505b67125ff7f85fa/locate.go#L75)):*

Is this default path relative to the root directory? It should probably be an absolute path to be on the safe side.
</blockquote></details>

I added the full path.
___
*[`limits/config.yaml` line 3 at r6](https://reviewable.io/reviews/m-lab/locate/169#-Nl0EPej4iwM13qgOr0u:-Nl0vaeB0uB4bKAAsCec:b4c9lz5) ([raw file](https://github.com/m-lab/locate/blob/28ef3a23c7095c98f271037505b67125ff7f85fa/limits/config.yaml#L3)):*

Just noting that this should catch the periodic client and any retries for up to one minute, but that we might need increase the duration if we find that the client is set to retry for a good while.
</blockquote></details>

:thumbsup:.

PR #188, opened 2024-05-16
Add -hostname-file command-line argument to read hostname from file.

robertodauria:
This PR allows setting the hostname via a file and the `-hostname-file` command-line flag. If both `-hostname` and `-hostname-file` are provided, `-hostname` takes precedence.

Note: I'm not super happy about the two different ways of providing paths, i.e. `hostname-file=/path/to/file` vs `registration-url=file:///path/to/file`. Using `flagx.URL` for both would help by making the syntax uniform, but using an URL for the hostname file seems quite wrong to me since it will always be a local path. Do you have any thoughts on this?

### DIFF
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -30,6 +30,7 @@ import (
 var (
 	heartbeatURL        string
 	hostname            string
+	hostnameFile        flagx.FileBytes
 	experiment          string
 	pod                 string
 	node                string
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
I think the `flagx.Bytes` type is appropriate.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/188#-:-Ny-uvJj4dMZuxkrPWw0:bqoe0zh)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on @robertodauria)

___
*[`cmd/heartbeat/main.go` line 33 at r1](https://reviewable.io/reviews/m-lab/locate/188#-Ny-ukZb5isMC-CsVrKC:-Ny-ukZb5isMC-CsVrKD:b-w9riyr) ([raw file](https://github.com/m-lab/locate/blob/313670fae84a20d5a0b69a70d90f4cb0f3f1fe12/cmd/heartbeat/main.go#L33)):*
> ```Go
> 	heartbeatURL        string
> 	hostname            string
> 	hostnameFile        flagx.FileBytes
> ```

This is such a cool flag!

PR #191, opened 2024-08-06
 Reads value of var lbPath (/metadata/loadbalanced)

nkinkade:
Previously, the file /metadata/loadbalanced only existed on MIG instances. We recently decided that the file should exist on every machine with either a "true" or "false" value. Locate was only checking for the file's existence to determine if the machine was loadbalanced or not, but since this is no longer a valid signal, this commit causes heartbeat to read the value of the file, and only treat the machine as loadbalanced if the value is "true".

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/191)

### DIFF
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -99,9 +99,7 @@ func main() {
 	var hc Checker
 
 	lbbytes, lberr := os.ReadFile(lbPath)
-	if lberr != nil {
-		rtx.Must(err, "failed to read the value of metadata file 'loadbalanced'")
-	}
+	rtx.Must(lberr, "failed to read metadata file 'loadbalanced'")
 
 	// If the "loadbalanced" file exists, then the instance is a load balanced VM.
 	// If not, then it is a standalone instance.
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -108,7 +108,7 @@ func main() {
 	// If the "loadbalanced" file exists, then make sure that the content of the
 	// file is "true". If the file doesn't exist, then, for now, just consider
 	// the machine as not loadbalanced.
-	if string(lbbytes) == "true" && lberr == nil {
+	if lberr == nil && string(lbbytes) == "true" {
 		gcpmd, err := metadata.NewGCPMetadata(md.NewClient(http.DefaultClient), hostname)
 		rtx.Must(err, "failed to get VM metadata")
 		gceClient, err := compute.NewRegionBackendServicesRESTClient(mainCtx)
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -94,14 +94,18 @@ func main() {
 	err = conn.Dial(heartbeatURL, http.Header{}, hbm)
 	rtx.Must(err, "failed to establish a websocket connection with %s", heartbeatURL)
 
-	_, lberr := os.ReadFile(lbPath)
 	probe := health.NewPortProbe(svcs)
 	ec := health.NewEndpointClient(static.HealthEndpointTimeout)
 	var hc Checker
 
+	lbbytes, lberr := os.ReadFile(lbPath)
+	if lberr != nil {
+		rtx.Must(err, "failed to read the value of metadata file 'loadbalanced'")
+	}
+
 	// If the "loadbalanced" file exists, then the instance is a load balanced VM.
 	// If not, then it is a standalone instance.
-	if lberr == nil {
+	if string(lbbytes) == "true" {
 		gcpmd, err := metadata.NewGCPMetadata(md.NewClient(http.DefaultClient), hostname)
 		rtx.Must(err, "failed to get VM metadata")
 		gceClient, err := compute.NewRegionBackendServicesRESTClient(mainCtx)
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -105,8 +105,9 @@ func main() {
 	// completes in 4 or 5 days, as of this comment 2024-08-06.
 	lbbytes, lberr := os.ReadFile(lbPath)
 
-	// If the "loadbalanced" file exists, then the instance is a load balanced VM.
-	// If not, then it is a standalone instance.
+	// If the "loadbalanced" file exists, then make sure that the content of the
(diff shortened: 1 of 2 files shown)


cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/191)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`cmd/heartbeat/main.go` line 102 at r1](https://reviewable.io/reviews/m-lab/locate/191#-O3dlKIP1ezB_7zx_a5l:-O3dlKIQ9fHICvCzSM_f:bxfujjb) ([raw file](https://github.com/m-lab/locate/blob/7172052ad7e692be931f597af19863b7fd5de4b6/cmd/heartbeat/main.go#L102)):*
> ```Go
> 
> 	lbbytes, lberr := os.ReadFile(lbPath)
> 	rtx.Must(lberr, "failed to read metadata file 'loadbalanced'")
> ```

To clarify, next time the heartbeat restarts the file will exist, right? Otherwise the pod will crash.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/191)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/main.go` line 102 at r1](https://reviewable.io/reviews/m-lab/locate/191#-O3dlKIP1ezB_7zx_a5l:-O3dltdk7q60L0MRbIUV:bmbu5gr) ([raw file](https://github.com/m-lab/locate/blob/7172052ad7e692be931f597af19863b7fd5de4b6/cmd/heartbeat/main.go#L102)):*

To clarify, next time the heartbeat restarts the file will exist, right? Otherwise the pod will crash.
</blockquote></details>

Good point. The file exists now on all VMs, but won't on physical machines until the rolling reboot completes. However, for every machine that reboots under the current conditions, it will have crashlooping pods, since heartbeat will think it is loadbalanced, even if it isn't, which is most instances.

Maybe what I can do in the meantime is to manually place that file on all production physical machines, then it should be safe to deploy this updated version of heartbeat before the rolling reboot has completed. What do you think?

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/191)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`cmd/heartbeat/main.go` line 102 at r1](https://reviewable.io/reviews/m-lab/locate/191#-O3dlKIP1ezB_7zx_a5l:-O3dmw_FA3Snqk3s7J0g:b-ly1wzf) ([raw file](https://github.com/m-lab/locate/blob/7172052ad7e692be931f597af19863b7fd5de4b6/cmd/heartbeat/main.go#L102)):*

Good point. The file exists now on all VMs, but won't on physical machines until the rolling reboot completes. However, for every machine that reboots under the current conditions, it will have crashlooping pods, since heartbeat will think it is loadbalanced, even if it isn't, which is most instances.

Maybe what I can do in the meantime is to manually place that file on all production physical machines, then it should be safe to deploy this updated version of heartbeat before the rolling reboot has completed. What do you think?
</blockquote></details>

We could also remove this failing condition for now and change the check below to something like `if lberr == nil && string(lbbytes) == "true")`. 

Adding the file manually could work, too, but I'd rather avoid manual changes in production, if possible.

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/191)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/main.go` line 102 at r1](https://reviewable.io/reviews/m-lab/locate/191#-O3dlKIP1ezB_7zx_a5l:-O3dqpHs7YQCmRnsC7mv:by7ixr0) ([raw file](https://github.com/m-lab/locate/blob/7172052ad7e692be931f597af19863b7fd5de4b6/cmd/heartbeat/main.go#L102)):*

We could also remove this failing condition for now and change the check below to something like `if lberr == nil && string(lbbytes) == "true")`. 

Adding the file manually could work, too, but I'd rather avoid manual changes in production, if possible.
</blockquote></details>

I added your suggestion. For now, heartbeat won't fail if the file doesn't exist.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/191)* status: 0 of 1 approvals obtained (waiting on @nkinkade)

___
*[`cmd/heartbeat/main.go` line 110 at r2](https://reviewable.io/reviews/m-lab/locate/191#-O3dsx0w0TbwEpTI7Wui:-O3dsx0w0TbwEpTI7Wuj:b-hdux7f) ([raw file](https://github.com/m-lab/locate/blob/d2bfdb4427442aed820ac293b7f0b081ed200d35/cmd/heartbeat/main.go#L110)):*
> ```Go
> 	// If the "loadbalanced" file exists, then the instance is a load balanced VM.
> 	// If not, then it is a standalone instance.
> 	if string(lbbytes) == "true" && lberr == nil {
> ```

For the cases where the file doesn't exist, should the order of the conditions be reversed? i.e., `if lberr == nil && string(lbbytes) == "true"`

nkinkade (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/191)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`cmd/heartbeat/main.go` line 110 at r2](https://reviewable.io/reviews/m-lab/locate/191#-O3dsx0w0TbwEpTI7Wui:-O3du1guCp33pleTtqAp:bhltvzg) ([raw file](https://github.com/m-lab/locate/blob/d2bfdb4427442aed820ac293b7f0b081ed200d35/cmd/heartbeat/main.go#L110)):*

For the cases where the file doesn't exist, should the order of the conditions be reversed? i.e., `if lberr == nil && string(lbbytes) == "true"`
</blockquote></details>

I reversed the order of the conditionals.

cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/191#-:-O3dv09vAVA2L9q5KbN_:bg7igkv)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #202, opened 2024-11-05
Always pick nodes when org filter is specified

stephen-soltesz:
This change adds an additional condition to the `alwaysPick` expression that always picks when the "Org" filter is specified.

This is necessary because some autonode deployments are deployed at less than probability 1, with a small enough set of sites that Locate results can return 500 errors.


robertodauria (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/202#-:-OAw1hwmAMRW-ISiJuwJ:bnfp4nl)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #203, opened 2024-11-05
Update Prometheus signals every time a Registration message is received

cristinaleonr:
This change is to avoid heartbeat instances from (re-)registering and having to wait up to 1min (Prometheus cronjob period) to get Prometheus health signals.

### DIFF
--- a/handler/heartbeat.go
+++ b/handler/heartbeat.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"context"
 	"encoding/json"
 	"net/http"
 	"time"
@@ -67,6 +68,12 @@ func (c *Client) handleHeartbeats(ws conn) error {
 					closeConnection(experiment, err)
 					return err
 				}
+
+				// Update Prometheus signals every time a Registration message is received.
+				if err := c.QueryPrometheus(context.Background()); err != nil {
+					log.Errorf("failed to query Prometheus, err: %v", err)
+				}
+
 				if hostname == "" {
 					hostname = hbm.Registration.Hostname
 					experiment = hbm.Registration.Experiment
--- a/handler/prometheus.go
+++ b/handler/prometheus.go
@@ -35,28 +35,37 @@ var (
 
 // Prometheus is a handler that collects Prometheus health signals.
 func (c *Client) Prometheus(rw http.ResponseWriter, req *http.Request) {
-	hostnames, err := c.query(req.Context(), e2eQuery, e2eLabel, e2eFunction)
+	err := c.QueryPrometheus(req.Context())
 	if err != nil {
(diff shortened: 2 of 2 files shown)


nkinkade (approved):
LGTM, with one possibly non-blocking comment.

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/203#-:-OB1nTSy9ajep7y1DGC3:byj2kti)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`handler/prometheus.go` line 109 at r2](https://reviewable.io/reviews/m-lab/locate/203#-OB1moZg8bFBfpeikFzf:-OB1moZg8bFBfpeikFzg:bslr5v5) ([raw file](https://github.com/m-lab/locate/blob/e21a1e6a028bfc048dd4925aa0032236f93efde4/handler/prometheus.go#L109)):*
> ```Go
> func formatQuery(query, filter string) string {
> 	if filter != "" {
> 		return fmt.Sprintf("%s{%s}", query, filter)
> ```

Just noting that this makes the assumption that  `query` will always just be a raw metric name. Today that is fine, since the gmx and e2e queries are just that. If in the future we added some new query that looked like `my_new_metric{status="ready"}`, that this would result in an invalid query like `my_new_metric{status="ready"}{machine="my-machine"}`.

cristinaleonr (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/203)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[`handler/prometheus.go` line 109 at r2](https://reviewable.io/reviews/m-lab/locate/203#-OB1moZg8bFBfpeikFzf:-OB1opA67t6ulnq-pVM_:bvrdl5c) ([raw file](https://github.com/m-lab/locate/blob/e21a1e6a028bfc048dd4925aa0032236f93efde4/handler/prometheus.go#L109)):*

Just noting that this makes the assumption that  `query` will always just be a raw metric name. Today that is fine, since the gmx and e2e queries are just that. If in the future we added some new query that looked like `my_new_metric{status="ready"}`, that this would result in an invalid query like `my_new_metric{status="ready"}{machine="my-machine"}`.
</blockquote></details>

Agreed. The raw queries should be unfiltered by design because they query metrics for all instances.

nkinkade (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/203)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #210, opened 2025-02-07
Add sketch package

robertodauria:
This PR adds a new package, `sketch`, with a Count-Min Sketch implementation backed by Redis. It's the first part of multiple PRs that will:

1. Use this data structure for rate limiting based on IP addresses or IP+user-agent
2. Add rate limiting to the Locate handlers

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/210)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.20
 require (
 	cloud.google.com/go/compute/metadata v0.2.3
 	cloud.google.com/go/secretmanager v1.11.2
+	github.com/alicebob/miniredis/v2 v2.34.0
 	github.com/apex/log v1.9.0
 	github.com/cenkalti/backoff/v4 v4.1.3
 	github.com/go-test/deep v1.0.8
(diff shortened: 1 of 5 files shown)


cristinaleonr (commented):
General question, why is a probabilistic data structure needed? In particular, given we're using Redis, I would expect a sliding window counter with sorted sets in Redis to result in a more precise and straightforward implementation.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/210#-:-OIXXNz22m9kMbXXxp-w:b-wyrz50)* status: 0 of 1 approvals obtained (waiting on @robertodauria)

___
*[`sketch/interface.go` line 9 at r2](https://reviewable.io/reviews/m-lab/locate/210#-OIXS67J79LT7usoA2lh:-OIXS67J79LT7usoA2li:b79iwf4) ([raw file](https://github.com/m-lab/locate/blob/38d882af953553c562e9844fc93a9ea18d38adae/sketch/interface.go#L9)):*
> Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values.

[Source](https://go.dev/wiki/CodeReviewComments#interfaces). Should the interface be in the consumer?
___
*[`sketch/countmin_test.go` line 12 at r2](https://reviewable.io/reviews/m-lab/locate/210#-OIXZ84U--LneovHVmCm:-OIXZ84V36WbiE6mDX6y:b-niyq1j) ([raw file](https://github.com/m-lab/locate/blob/38d882af953553c562e9844fc93a9ea18d38adae/sketch/countmin_test.go#L12)):*
> ```Go
> )
> 
> func setupTestRedis(t *testing.T) (*redis.Pool, func()) {
> ```

This is helpful :thumbsup:.

robertodauria (commented):
I expanded a bit more on Slack, but this was my original reasoning:
- Memory usage: a count-min sketch size does not depend on how many entries are added
- It's relatively easy for an abusive client to randomize the UA, not only making the rate limiting useless but also causing an explosion in the number of keys
- The individual Redis operations used in this implementation are O(1) -- turns out this is misleading, see below

As I found after re-implementing this with ZSETs and benchmarking both solutions:
- This mapping of the CM Sketch semantics on Redis hashmaps is quite inefficient (with depth=7, we need 7 operations to increment and 7 operations to count)
- The memory usage benefit isn't really there unless the number of keys is in the millions.
- ZSETs are drastically faster (~4x) and the memory usage is actually lower for the scale we operate at. We likely won't have enough entries to start seeing the benefits of a CM sketch.

I would probably still advocate for a CM Sketch if we had access to Redis Enterprise, which provides a native implementation written in C with O(1) operations and much lower memory footprint, but this is not the case.

Considering the above, I'm closing this PR to propose the ZSET-based implementation instead. Thanks!

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/210#-:-OIgruba4me53_TSGcVk:bhdq1h4)* status: 0 of 1 approvals obtained

PR #211, opened 2025-02-09
Implement rate limiting based on IP+UA

robertodauria:
Implements a rate limiter backed by Redis' sorted sets (ZSETs).

```
// It maintains a sliding window of events for each IP+UA combination, where:
//   - Each event is stored in a ZSET with the timestamp as score
//   - Old events (outside the window) are automatically removed
//   - Keys automatically expire after the configured interval
//
// The limiter considers a request to be rate-limited if the number of events
// in the current window exceeds MaxEvents.
```

Initially we are only going to log which IP+UA are being rate limited and increase a Prometheus counter. Once we confirm that the limiter is WAI, we can start actually rejecting requests.

I used a separate Redis instance because:
- The memorystore package appears to assume its Redis instance won't be used for anything else and expects all keys to be hashes (with a specific structure)
- I don't think we should create a single point of failure. If, for any reason, the memory usage of the rate limiter spikes and Redis can no longer handle queries, Locate will act conservatively by allowing the client's request to pass through. This way, the damage is minimized. However, we cannot afford to lose the heartbeat registrations.

Supersedes https://github.com/m-lab/locate/pull/210

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/211)

### DIFF
--- a/cloudbuild/app.yaml.mlab-ns.template
+++ b/cloudbuild/app.yaml.mlab-ns.template
@@ -37,5 +37,6 @@ env_variables:
   LOCATOR_MAXMIND: true
   MAXMIND_URL: gs://downloader-{{PLATFORM_PROJECT}}/Maxmind/current/GeoLite2-City.tar.gz
   REDIS_ADDRESS: {{REDIS_ADDRESS}}
+  RATE_LIMIT_REDIS_ADDRESS: {{RATE_LIMIT_REDIS_ADDRESS}}
   PROMETHEUSX_LISTEN_ADDRESS: ':9090' # Must match one of the forwarded_ports above.
   PROMETHEUS_URL: 'https://prometheus-basicauth.{{PLATFORM_PROJECT}}.measurementlab.net/'
--- a/cloudbuild/app.yaml.template
+++ b/cloudbuild/app.yaml.template
@@ -36,5 +36,6 @@ env_variables:
   LOCATOR_MAXMIND: true
   MAXMIND_URL: gs://downloader-{{PLATFORM_PROJECT}}/Maxmind/current/GeoLite2-City.tar.gz
   REDIS_ADDRESS: {{REDIS_ADDRESS}}
+  RATE_LIMIT_REDIS_ADDRESS: {{RATE_LIMIT_REDIS_ADDRESS}}
   PROMETHEUSX_LISTEN_ADDRESS: ':9090' # Must match one of the forwarded_ports above.
   PROMETHEUS_URL: 'https://prometheus-basicauth.{{PLATFORM_PROJECT}}.measurementlab.net/'
--- a/cloudbuild/cloudbuild.yaml
+++ b/cloudbuild/cloudbuild.yaml
@@ -2,55 +2,55 @@ timeout: 3600s
 
 options:
   env:
-    - PROJECT_ID=$PROJECT_ID
-    - WORKSPACE_LINK=/go/src/github.com/m-lab/locate
+  - PROJECT_ID=$PROJECT_ID
+  - WORKSPACE_LINK=/go/src/github.com/m-lab/locate
 
 steps:
-  # Run unit tests for environment.
-  - name: gcr.io/$PROJECT_ID/golang-cbif:1.20
-    env:
-      - GONOPROXY=github.com/m-lab/go/*
-    args:
-      - go version
-      - go get -v -t ./...
-      - go vet ./...
-      - go test ./... -race
-      - go test -v ./...
+# Run unit tests for environment.
+- name: gcr.io/$PROJECT_ID/golang-cbif:1.20
+  env:
+  - GONOPROXY=github.com/m-lab/go/*
+  args:
+  - go version
+  - go get -v -t ./...
+  - go vet ./...
+  - go test ./... -race
+  - go test -v ./...
 
-  # Deployment of APIs in sandbox & staging.
-  - name: gcr.io/$PROJECT_ID/gcloud-jsonnet-cbif:1.1
-    env:
-      # Use cbif condition: only run these steps in one of these projects.
-      - PROJECT_IN=mlab-sandbox,mlab-staging
-    args:
-      - cp cloudbuild/app.yaml.template app.yaml
-      - >
-        sed -i
-        -e 's/{{PROJECT}}/$PROJECT_ID/g'
-        -e 's/{{PLATFORM_PROJECT}}/$_PLATFORM_PROJECT/'
-        -e 's/{{REDIS_ADDRESS}}/$_REDIS_ADDRESS/'
-        -e 's/{{RATE_LIMIT_REDIS_ADDRESS}}/$_RATE_LIMIT_REDIS_ADDRESS/'
-        app.yaml
-      - gcloud --project $PROJECT_ID app deploy --promote app.yaml
-      # After deploying the new service, deploy the openapi spec.
-      - sed -i -e 's/{{PROJECT}}/$PROJECT_ID/' -e 's/{{DEPLOYMENT}}/$PROJECT_ID/' openapi.yaml
-      - gcloud endpoints services deploy openapi.yaml
+# Deployment of APIs in sandbox & staging.
+- name: gcr.io/$PROJECT_ID/gcloud-jsonnet-cbif:1.1
+  env:
+  # Use cbif condition: only run these steps in one of these projects.
+  - PROJECT_IN=mlab-sandbox,mlab-staging
+  args:
+  - cp cloudbuild/app.yaml.template app.yaml
+  - >
+    sed -i
+    -e 's/{{PROJECT}}/$PROJECT_ID/g'
+    -e 's/{{PLATFORM_PROJECT}}/$_PLATFORM_PROJECT/'
+    -e 's/{{REDIS_ADDRESS}}/$_REDIS_ADDRESS/'
+    -e 's/{{RATE_LIMIT_REDIS_ADDRESS}}/$_RATE_LIMIT_REDIS_ADDRESS/'
+    app.yaml
+  - gcloud --project $PROJECT_ID app deploy --promote app.yaml
+  # After deploying the new service, deploy the openapi spec.
+  - sed -i -e 's/{{PROJECT}}/$PROJECT_ID/' -e 's/{{DEPLOYMENT}}/$PROJECT_ID/' openapi.yaml
+  - gcloud endpoints services deploy openapi.yaml
 
-  # Deployment of APIs in mlab-ns.
-  - name: gcr.io/$PROJECT_ID/gcloud-jsonnet-cbif:1.1
-    env:
-      # Use cbif condition: only run these steps in this project.
-      - PROJECT_IN=mlab-ns
-    args:
-      - cp cloudbuild/app.yaml.mlab-ns.template app.yaml
-      - >
-        sed -i
-        -e 's/{{PROJECT}}/$PROJECT_ID/g'
-        -e 's/{{PLATFORM_PROJECT}}/$_PLATFORM_PROJECT/'
-        -e 's/{{REDIS_ADDRESS}}/$_REDIS_ADDRESS/'
-        -e 's/{{RATE_LIMIT_REDIS_ADDRESS}}/$_RATE_LIMIT_REDIS_ADDRESS/'
-        app.yaml
-      - gcloud --project $PROJECT_ID app deploy --promote app.yaml
-      # After deploying the new service, deploy the openapi spec.
-      - sed -i -e 's/{{PROJECT}}/$PROJECT_ID/' -e 's/{{DEPLOYMENT}}/Production/' openapi.yaml
-      - gcloud endpoints services deploy openapi.yaml
+# Deployment of APIs in mlab-ns.
+- name: gcr.io/$PROJECT_ID/gcloud-jsonnet-cbif:1.1
+  env:
+  # Use cbif condition: only run these steps in this project.
+  - PROJECT_IN=mlab-ns
+  args:
+  - cp cloudbuild/app.yaml.mlab-ns.template app.yaml
+  - >
+    sed -i
+    -e 's/{{PROJECT}}/$PROJECT_ID/g'
+    -e 's/{{PLATFORM_PROJECT}}/$_PLATFORM_PROJECT/'
+    -e 's/{{REDIS_ADDRESS}}/$_REDIS_ADDRESS/'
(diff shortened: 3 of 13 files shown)


cristinaleonr (commented):
This is great, @robertodauria! Thank you!

> Reviewed 1 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/211#-:-OIkfTw54nenypAt6Yhw:bj2pfwe)* status: 0 of 1 approvals obtained

___
*[`locate.go` line 82 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkVY9OCtvuDxtT-7am:-OIkVY9OCtvuDxtT-7an:brvomkf) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/locate.go#L82)):*
> ```Go
> 	flag.DurationVar(&rateLimitInterval, "rate-limit-interval", time.Hour, "Time window for IP+UA rate limiting")
> 	flag.IntVar(&rateLimitMax, "rate-limit-max", 40, "Max number of events in the time window for IP+UA rate limiting")
> 	flag.StringVar(&rateLimitPrefix, "rate-limit-prefix", "locate:ratelimit:", "Prefix for Redis keys for IP+UA rate limiting")
> ```

What do we need the prefix for?
___
*[`handler/handler.go` line 166 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkedNr1B51XQBTYTIP:-OIkedNr1B51XQBTYTIQ:b-7vdkq5) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/handler/handler.go#L166)):*
> ```Go
> 		ips := strings.Split(ip, ",")
> 		ip = strings.TrimSpace(ips[0])
> 		if ip != "" {
> ```

Given this is a critical path, I'd rather guarantee it does not panic, even if it shouldn't because of App Engine. So, I would suggest something like.

`if len(IPs) > 0) {`
`   ip = strings.TrimSpace(ips[0])`

And the rate limiter can do its own validation regarding whether the IP is valid/non-empty and return an error if so.



_Code quote:_
```Go
		ip = strings.TrimSpace(ips[0])
		if ip != "" {
```
___
*[`handler/handler.go` line 172 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkWANK8FuDyt8nuYzd:-OIkWANK8FuDyt8nuYze:b-n17dta) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/handler/handler.go#L172)):*
> ```Go
> 			limited, err := c.ipLimiter.IsLimited(ip, ua)
> 			if err != nil {
> 				// Log error but don't block request (fail open).
> ```

Maybe we can add a TODO to add testing for this path when we switch to fail closed.
___
*[`limits/ratelimiter_bench_test.go` line 61 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkZ_lQ9sJRCf1cyfXI:-OIkZ_lQ9sJRCf1cyfXJ:b-2fj5mh) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/limits/ratelimiter_bench_test.go#L61)):*
> ```Go
> 	tests := []struct {
> 		name     string
> 		duration time.Duration
> ```

Is duration used?
___
*[`limits/ratelimiter.go` line 11 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkXYfc2nOHCmqoGMj2:-OIkXYfc2nOHCmqoGMj3:b7y9c0q) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/limits/ratelimiter.go#L11)):*
> ```Go
> )
> 
> // RateLimitConfig holds the configuration for IP+UA rate limiting
> ```

I really like this config.
___
*[`limits/ratelimiter.go` line 25 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkZykn8DmvM_hBH-wf:-OIkZykn8DmvM_hBH-wg:b4c9lz5) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/limits/ratelimiter.go#L25)):*
> ```Go
> //
> // The limiter considers a request to be rate-limited if the number of events
> // in the current window exceeds MaxEvents.
> ```

:thumbsup:.



_Code quote:_
```Go
// RateLimiter implements a distributed rate limiter using Redis sorted sets (ZSET).
// It maintains a sliding window of events for each IP+UA combination, where:
//   - Each event is stored in a ZSET with the timestamp as score
//   - Old events (outside the window) are automatically removed
//   - Keys automatically expire after the configured interval
//
// The limiter considers a request to be rate-limited if the number of events
// in the current window exceeds MaxEvents.
```
___
*[`limits/ratelimiter.go` line 34 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkXebq9HbJTm1_HrRO:-OIkXebq9HbJTm1_HrRP:b68zagi) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/limits/ratelimiter.go#L34)):*
> ```Go
> 
> // NewRateLimiter creates a new rate limiter
> func NewRateLimiter(pool *redis.Pool, config RateLimitConfig) *RateLimiter {
> ```

*Nit*: [Go comments should end in a period](https://go.dev/wiki/CodeReviewComments#comment-sentences).
___
*[`limits/ratelimiter.go` line 53 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkd6KME5H2aUTJ6hYT:-OIkd6KME5H2aUTJ6hYU:bqlpk71) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/limits/ratelimiter.go#L53)):*
> ```Go
> 	defer conn.Close()
> 
> 	now := time.Now().UnixMicro()
> ```

Should this be `UTC` to guarantee consistency?

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/211)* status: 0 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`locate.go` line 82 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkVY9OCtvuDxtT-7am:-OIl6k9H5NKVpK-xuZsa:b-1mbwc6) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/locate.go#L82)):*

What do we need the prefix for?
</blockquote></details>

Hmm. I originally added it when I still hadn't switched to a separate Redis instance dedicated to rate limiting. I guess since nothing else will go on the same Redis it's not as useful anymore. However, in case we ever store other (ratelimiting-related) keys there it'll be useful, and adding a prefix later when the system is working in production is a bit more annoying. What do you think? Perhaps I can just make it a constant?
___
*[`handler/handler.go` line 166 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkedNr1B51XQBTYTIP:-OIl6ADW68t4w0u3r1E_:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/handler/handler.go#L166)):*

Given this is a critical path, I'd rather guarantee it does not panic, even if it shouldn't because of App Engine. So, I would suggest something like.

`if len(IPs) > 0) {`
`   ip = strings.TrimSpace(ips[0])`

And the rate limiter can do its own validation regarding whether the IP is valid/non-empty and return an error if so.
</blockquote></details>

Done.
___
*[`handler/handler.go` line 172 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkWANK8FuDyt8nuYzd:-OIl6EXTCtSdM13EoInd:bju12w0) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/handler/handler.go#L172)):*

Maybe we can add a TODO to add testing for this path when we switch to fail closed.
</blockquote></details>

In this particular path I don't think we'll ever want to fail open -- if Redis is down for any reason and IsLimited returns an error I'd rather have all the requests passing than all of them blocked.
However, I added a TODO in the `else if limited` below.
___
*[`limits/ratelimiter.go` line 11 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkXYfc2nOHCmqoGMj2:-OIl3zn68tEVrNcToELl:b32vhw) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/limits/ratelimiter.go#L11)):*

I really like this config.
</blockquote></details>

:)
___
*[`limits/ratelimiter.go` line 34 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkXebq9HbJTm1_HrRO:-OIl58C4BylZKfYR38B5:b-896fix) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/limits/ratelimiter.go#L34)):*

*Nit*: [Go comments should end in a period](https://go.dev/wiki/CodeReviewComments#comment-sentences).
</blockquote></details>

Done.
___
*[`limits/ratelimiter.go` line 53 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkd6KME5H2aUTJ6hYT:-OIl4WId-qn6cNF3Hqma:bud4fkf) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/limits/ratelimiter.go#L53)):*

Should this be `UTC` to guarantee consistency?
</blockquote></details>

(discussed elsewhere) I confirmed that Time.UnixMicro() does not depend on the location, i.e. it's always UTC.
___
*[`limits/ratelimiter_bench_test.go` line 61 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkZ_lQ9sJRCf1cyfXI:-OIl4pLcBDyqph2SGQ1O:bit209q) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/limits/ratelimiter_bench_test.go#L61)):*

Is duration used?
</blockquote></details>

Done. I changed this benchmark so many times that I forgot that now unused parameter in.

cristinaleonr (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/211#-:-OIlBtgx1eh_kOaH1wBY:bg7igkv)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on @robertodauria)

___
*[`locate.go` line 82 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkVY9OCtvuDxtT-7am:-OIlBUZn6oLGu1Qh1hMk:b59vegi) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/locate.go#L82)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

Hmm. I originally added it when I still hadn't switched to a separate Redis instance dedicated to rate limiting. I guess since nothing else will go on the same Redis it's not as useful anymore. However, in case we ever store other (ratelimiting-related) keys there it'll be useful, and adding a prefix later when the system is working in production is a bit more annoying. What do you think? Perhaps I can just make it a constant?
</blockquote></details>

Having it in case of future extensions makes sense to me :thumbsup:.
___
*[`handler/handler.go` line 172 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkWANK8FuDyt8nuYzd:-OIl9zR23qHXAZmzdhvY:b2ayt24) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/handler/handler.go#L172)):*
<details><summary><i>Previously, robertodauria (Roberto D'Auria) wrote…</i></summary><blockquote>

In this particular path I don't think we'll ever want to fail open -- if Redis is down for any reason and IsLimited returns an error I'd rather have all the requests passing than all of them blocked.
However, I added a TODO in the `else if limited` below.
</blockquote></details>

I assume you mean you don't think we'll ever want to fail *closed*.

Thanks for adding the comment. The main point I was trying to convey is that we should remember to add some unit testing in the handler for this new functionality before we enable the request blocking.
___
*[`limits/ratelimiter.go` line 48 at r2](https://reviewable.io/reviews/m-lab/locate/211#-OIlAymC51N4j7rUY4Zl:-OIlAymC51N4j7rUY4Zm:b6edd23) ([raw file](https://github.com/m-lab/locate/blob/c2adcc090316733672648c8ea4f02ba0092ebd04/limits/ratelimiter.go#L48)):*
> ```Go
> // generateKey creates a Redis key from IP and User-Agent.
> func (rl *RateLimiter) generateKey(ip, ua string) string {
> 	return fmt.Sprintf("%s%s:%s", rl.keyPrefix, ip, ua)
> ```

I would suggest that this be `"%s:%s:%s"` and we remove the colon from the prefix. That way the prefix can be changed/others can be used and we don't have to remember to always add the colon.



_Code quote:_
```Go
"%s%s:%s"
```

robertodauria (commented):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/211)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on @cristinaleonr)

___
*[`handler/handler.go` line 172 at r1](https://reviewable.io/reviews/m-lab/locate/211#-OIkWANK8FuDyt8nuYzd:-OIlJxLr5D8Np-PPDTMO:b-7jbmcf) ([raw file](https://github.com/m-lab/locate/blob/92c85132c904674376a209cf6a6162c6dc90fbdc/handler/handler.go#L172)):*

I assume you mean you don't think we'll ever want to fail *closed*.

Thanks for adding the comment. The main point I was trying to convey is that we should remember to add some unit testing in the handler for this new functionality before we enable the request blocking.
</blockquote></details>

Yup. I had edited the comment on GitHub, but Reviewable doesn't seem to pick up edits. I meant we don't ever want to fail *closed*. And, I've added another TODO to test the path in any case.
___
*[`limits/ratelimiter.go` line 48 at r2](https://reviewable.io/reviews/m-lab/locate/211#-OIlAymC51N4j7rUY4Zl:-OIlNSQE1UiWAPUjljcH:boyexbg) ([raw file](https://github.com/m-lab/locate/blob/c2adcc090316733672648c8ea4f02ba0092ebd04/limits/ratelimiter.go#L48)):*

I would suggest that this be `"%s:%s:%s"` and we remove the colon from the prefix. That way the prefix can be changed/others can be used and we don't have to remember to always add the colon.
</blockquote></details>

Done. Thanks.

PR #212, opened 2025-02-10
Fix nil pointer in Nearest handler

robertodauria:
When I changed the handler to not actually block I kept a reference to `result` that's nil at that point.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/212)

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -176,7 +176,7 @@ func (c *Client) Nearest(rw http.ResponseWriter, req *http.Request) {
 				log.Printf("Rate limiter error: %v", err)
 			} else if limited {
 				metrics.RequestsTotal.WithLabelValues("nearest", "rate limit",
-					http.StatusText(result.Error.Status)).Inc()
+					http.StatusText(http.StatusTooManyRequests)).Inc()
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/212)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on @nkinkade)

___
*[`handler/handler.go` line 173 at r1](https://reviewable.io/reviews/m-lab/locate/212#-OIla6AC12pDWpPJ3pk2:-OIla6AC12pDWpPJ3pk3:byuj9su) ([raw file](https://github.com/m-lab/locate/blob/4acbf0c3da726393fb878b875edd02adab24f87f/handler/handler.go#L173)):*
> ```Go
> 			ua := req.Header.Get("User-Agent")
> 			limited, err := c.ipLimiter.IsLimited(ip, ua)
> 			if err != nil {
> ```

Not strictly related to this PR, but at some point it would be good to add metrics to see how often we're hitting error paths.

PR #219, opened 2025-03-10
Update rate limiter to have different configurable limits for IP-only and IP+UA

robertodauria:
This allows to have different per-IP and per-IP+UA intervals and thresholds. If the IP address made more than X requests in Y time, the IP+UA key is ignored.

Adds the `-rate-limit-ip-interval` and `-rate-limit-ip-max` to configure this new rate limiting rule.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/219)

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -44,7 +44,7 @@ type Signer interface {
 }
 
 type Limiter interface {
-	IsLimited(ip, ua string) (bool, error)
+	IsLimited(ip, ua string) (limits.LimitStatus, error)
 }
 
 // Client contains state needed for xyz.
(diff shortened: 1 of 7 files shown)


cristinaleonr (approved):
It's good to test this out before enabling it 100% and try to get a sense of how much it would affect large integrations.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/219#-:-OL0i4rI-vQ5ZYFbFfaQ:b-kuydi9)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

robertodauria (commented):
I'm going to deploy this with 10% of the traffic, making sure to split traffic by IP address rather than random so that the same IP always goes to the same version and we get a realistic view.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/219#-:-OL4K85a7VkpvBdZElwW:bfjve2k)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #220, opened 2025-03-10
Always pass country/site/org/strict/machine-type

robertodauria:
This is part of the launch checklist for BYOS: all parameters that can influence Locate's load balancing decisions should be visibile in the data. This PR makes sure they are always passed along to ndt-server.

(Noting that we have no way to prevent the client from rewriting the URLs, but that's a different problem. This is enough for these particular qs parameters.)

### DIFF
--- a/static/configs.go
+++ b/static/configs.go
@@ -32,6 +32,11 @@ const (
 	EarlyExitParameter         = "early_exit"
 	MaxCwndGainParameter       = "max_cwnd_gain"
 	MaxElapsedTimeParameter    = "max_elapsed_time"
+	CountryParameter           = "country"
+	SiteParameter              = "site"
+	StrictParameter            = "strict"
+	OrgParameter               = "org"
+	MachineTypeParameter       = "machine-type"
 )
 
 // URL creates inline url.URLs.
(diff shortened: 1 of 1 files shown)


cristinaleonr (approved):
Is there any concern of the URLs exceeding a length limit?

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/220#-:-OL0kRzU5GQKojv0jz33:bg7l8r0)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

robertodauria (commented):
The limit used to be ~2000 characters back in 2009 and might still be that for some old browsers. Currently at least 8k characters seems to be supported everywhere, and the [HTTP semantics RFC](https://www.rfc-editor.org/rfc/rfc9110#section-4.1-5) now recommends 8k *octets* as the minimum that should be supported.

In any case, our current Locate URLs are ~520 characters.

> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/220#-:-OL4IRWQBfwDIXACxh4_:b-8o7x16)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained

PR #228, opened 2025-07-15
feat(auth): Add JWT authentication for heartbeat endpoints

robertodauria:
## Summary
- Adds JWT-protected `/v2/platform/heartbeat-jwt` endpoint with organization validation
- Updates heartbeat client to use JWT authentication with automatic token refresh
- Maintains backward compatibility with existing API key endpoint

### DIFF
--- a/cmd/heartbeat/main.go
+++ b/cmd/heartbeat/main.go
@@ -154,6 +154,13 @@ func main() {
 
 	// Establish a connection with JWT authentication.
 	conn := connection.NewConn()
+	
+	// Set up JWT token refresh for automatic token renewal
+	conn.SetTokenRefresher(func() (string, error) {
+		log.Printf("Refreshing JWT token...")
+		return getJWTTokenFunc(apiKey, tokenExchangeURL)
+	})
+	
 	err = conn.Dial(heartbeatURL, headers, hbm)
 	rtx.Must(err, "failed to establish a websocket connection with %s", heartbeatURL)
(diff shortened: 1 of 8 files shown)


nkinkade (approved):
LGTM, with one small comment about possibly using table driven tests.

> @nkinkade reviewed 5 of 8 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/228#-:-OYHaOYN4r7vYRbNU_zR:bqsdj78)* status: <span>:shipit:</span> complete! 1 of 1 approvals obtained (waiting on @bassosimone)

___
*[`connection/connection_test.go` line 203 at r2](https://reviewable.io/reviews/m-lab/locate/228#-OYHXfg1Bx3i_U0Q52Kw:-OYHXfg1Bx3i_U0Q52Kx:b2a3n96) ([raw file](https://github.com/m-lab/locate/blob/43813ef0fb6201756de1c9371b80a1a5b172b838/connection/connection_test.go#L203)):*
> ```Go
> }
> 
> func Test_TokenRefresh(t *testing.T) {
> ```

Could these three tests be converted into a single table driven test? They all seems pretty similar, but with perhaps differing tokens? I ask only because it seems like this should be possible, and my understanding is that table driven tests are preferred. That said, these tests seem fine. Same for a couple of the tests in heartbeat_test.go.

PR #229, opened 2025-10-23
fix: use PLATFORM_PROJECT for the auth service in openapi.yaml

robertodauria:
Use `_PLATFORM_PROJECT` instead of `PROJECT_ID` in OpenAPI configuration to handle deployment scenarios where the auth service runs in a different GCP project (e.g., mlab-ns deployment where auth service is in mlab-oti).

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/229)


nkinkade (approved):
LGTM

PR #230, opened 2025-11-18
feat: improve support for local testing

robertodauria:
This PR originates from the realization that running Locate locally and attaching heartbeat instances to it became significantly harder after introducing JWTs for authentication.

Here I implemented 3 selectable backends to verify JWTs:

1. `espv1`: (default) this is the previous behavior. It trusts the `X-Endpoint-API-UserInfo` header added/overwritten by Google App Engine's ESPv1 proxy, which contains the claims if and only if the `Authorization` header contained a valid JWT with a valid signature.
2. `direct`: allows Locate to perform its own JWT signature verification. Requires a `-jwt-jwks-url` flag pointing to a valid keyset to use for verification. Only for development usage.
3. `insecure`: does not verify JWTs and trusts them blindly. Only for development usage.

Additionally, since heartbeat needs to obtain a signed JWT to call Locate and we don't want the token-exchange dependency for local testing, I added a `-jwt-token` flag that allows to tell heartbeat which JWT to use directly.

Putting everything together:
```bash
# Start Redis + Locate
docker compose up --build

# Generate an unsigned JWT
export DEV_JWT=$(./scripts/create-dev-jwt.sh)

# Pass the generated JWT to heartbeat
go run ./cmd/heartbeat/main.go \
    -heartbeat-url ws://localhost:8080/v2/platform/heartbeat \
    -hostname=mlab1-lga0t.mlab-sandbox.measurement-lab.org \
    -experiment=ndt \
    -registration-url=https://siteinfo.mlab-oti.measurementlab.net/v2/sites/registration.json \
    -services 'ndt/ndt7=ws:///ndt/v7/download,ws:///ndt/v7/upload' \
    -jwt-token="$DEV_JWT"
    
# Listen on ports 80/443 so heartbeat reports a healthy node
sudo nc -l -k -p 80 &
sudo nc -l -k -p 443 &

# Wait for ~10s, then test
curl 'http://localhost:8080/v2/nearest/ndt/ndt7?region=US-NJ'
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/230)

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -114,6 +114,7 @@ func NewClient(project string, private Signer, locatorV2 LocatorV2, client Clien
 }
 
 // NewClientDirect creates a new client with a target template using only the target machine.
+// TODO: Remove this and use NewClient in test code.
 func NewClientDirect(project string, private Signer, locatorV2 LocatorV2, client ClientLocator, prom PrometheusClient) *Client {
 	return &Client{
 		Signer:           private,


handler/handler.go
bassosimone: ⚠️ Do we need to also assign the JWT verifier in this case?

Usually, in Go, I have always used a single constructor for each piece of code and tried to stay away from multiple constructors. Is using more than a single constructor per type common in M-Lab code? If so, I suspect I definitely have to keep this in mind when doing code reviews.

I also do wonder which is the use case for which this constructor is required. It seems a pretty niche use case, unless I misunderstand what is the actual objective here.

handler/handler_test.go
```diff
@@ -338,7 +338,7 @@ func TestClient_Nearest(t *testing.T) {
 			if tt.cl == nil {
 				tt.cl = clientgeo.NewAppEngineLocator()
 			}
-			c := NewClient(tt.project, tt.signer, tt.locator, tt.cl, prom.NewAPI(nil), tt.limits, tt.ipLimiter, nil)
+			c := NewClient(tt.project, tt.signer, tt.locator, tt.cl, prom.NewAPI(nil), tt.limits, tt.ipLimiter, nil, nil)
```
bassosimone: 🤔 I am not super happy about clearing the path for passing `nil` to types that should not be `nil` even for testing. Perhaps, you can just instantiate the simplest possible verifier?

auth/jwtverifier/insecure.go
```diff
@@ -0,0 +1,83 @@
...
+	warnedOnce sync.Once
+}
+
+// NewInsecureVerifier creates a new insecure JWT verifier.
+// Returns an error if the ALLOW_INSECURE_JWT environment variable is not set to "true".
+func NewInsecureVerifier() (*InsecureVerifier, error) {
+	// Require explicit environment variable as safety check
+	if os.Getenv("ALLOW_INSECURE_JWT") != "true" {
+		return nil, fmt.Errorf("insecure JWT mode requires ALLOW_INSECURE_JWT=true environment variable")
+	}
```
bassosimone: ❤️

auth/jwtverifier/insecure.go
```diff
@@ -0,0 +1,83 @@
...
+	authHeader := req.Header.Get("Authorization")
+	if authHeader == "" {
+		return nil, fmt.Errorf("Authorization header not found")
+	}
+
+	if !strings.HasPrefix(authHeader, "Bearer ") {
+		return nil, fmt.Errorf("Authorization header must be in format: Bearer <token>")
+	}
+
+	tokenString := strings.TrimPrefix(authHeader, "Bearer ")
```
bassosimone: Same comment as before for `strings.CutPrefix`.

auth/jwtverifier/insecure.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// Parse JWT without verification
```
bassosimone: 🤔 The comment says "Parse JWT without verification" but the code is the same as for `direct.go` and invokes the same `jwt.ParseSigned` function. So, I think it's a bit misleading. I think the part where you lack verification is where you extract claims, which is clearly marked below as insecure (good!).

auth/jwtverifier/verifier_test.go
```diff
@@ -0,0 +1,389 @@
...
+	if mode := verifier.Mode(); mode != "direct" {
+		t.Errorf("Mode() = %v, want direct", mode)
+	}
+}
+
+// Test InsecureVerifier
+func TestInsecureVerifier_ExtractClaims(t *testing.T) {
+	// Set environment variable for insecure mode
+	os.Setenv("ALLOW_INSECURE_JWT", "true")
+	defer os.Unsetenv("ALLOW_INSECURE_JWT")
```
bassosimone: 🗒️ FYI (and not necessarily related to the correctness of this pull request) `os.Setenv` and `os.Unsetenv` are thread unsafe operations that can lead to memory corruption when used from different thread contexts. Earlier this year, I had this issue in a pet project of mine here: https://github.com/rbmk-project/rbmk/pull/71.

bassosimone (changes requested):
Overall: I think it's very useful to make `m-lab/locate` work locally for testing. This feels like a very good quality of life improvements for developers.

This code review raises a big concern about the XFF header handling documentation or implementation, plus additional concerns and suggestions that you can read below.

Let's discuss!

auth/jwtverifier/verifier.go
```diff
@@ -0,0 +1,22 @@
...
+//   - Insecure: Parse JWT without validation (development/testing only)
+package jwtverifier
+
+import (
+	"net/http"
+)
+
+// JWTVerifier defines the interface for extracting JWT claims from HTTP requests.
+// Different implementations support different verification modes.
+type JWTVerifier interface {
```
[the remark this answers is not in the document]
robertodauria: Done.

auth/jwtverifier/verifier_test.go
```diff
@@ -0,0 +1,389 @@
+package jwtverifier
```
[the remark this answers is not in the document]
robertodauria: Done.

auth/jwtverifier/verifier_test.go
```diff
@@ -0,0 +1,389 @@
...
+			wantErr:     true,
+		},
+		{
+			name:        "invalid base64",
+			headerValue: "not-base64!@#$",
+			wantErr:     true,
+		},
+		{
+			name:        "invalid JSON",
+			headerValue: base64.StdEncoding.EncodeToString([]byte("not json")),
```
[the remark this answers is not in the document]
robertodauria: Discussed offline: `'not json'` is valid, `not json` is not json :slightly_smiling_face:

robertodauria (commented):
> @robertodauria made 9 comments.
> *[Reviewable](https://reviewable.io/reviews/m-lab/locate/230)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @bassosimone).

PR #255, opened 2026-03-04
feat: import dispatch.yaml from mlab-ns and deploy it via cloudbuild

robertodauria:
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/255)

### DIFF
--- a/cloudbuild/cloudbuild.yaml
+++ b/cloudbuild/cloudbuild.yaml
@@ -62,6 +62,7 @@ steps:
         -e 's/{{EARLY_EXIT_CLIENTS}}/$_EARLY_EXIT_CLIENTS/'
         app.yaml
       - gcloud --project $PROJECT_ID app deploy --no-promote app.yaml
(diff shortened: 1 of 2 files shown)


bassosimone (changes requested):
I think we should also deploy `dispatch.yaml` for `mlab-sandbox` and `mlab-staging` in the `cloudbuild.yaml` file, not just for `mlab-ns`. Otherwise we're not fully replicating what `m-lab/mlab-ns`'s `cloudbuild.yaml` used to do. It deployed `dispatch.yaml` to all three projects (`PROJECT_IN=mlab-sandbox,mlab-staging,mlab-ns` - see https://github.com/m-lab/mlab-ns/blob/v2023.02.13/cloudbuild.yaml#L29).

bassosimone (changes requested):
Also, it seems we are not ready yet to deploy this beyond staging. This merge request should be stalled until we have further operational clarity.

PR #260, opened 2026-04-20
feat(handler): propagate int_id/key_id into priority access tokens

robertodauria:
## Summary
- `/v2/priority/nearest/` now extracts both `int_id` and `key_id` from the integration JWT and embeds them into the resulting access token, so downstream experiments (e.g. ndt-server) can attribute measurements. Non-priority requests are unaffected.

## Do not merge yet
This PR pins `m-lab/access` to a commit (`0a3b3b6`) that currently lives only on a feature branch of `m-lab/access`, not on its `main`. Merge order should be:
1. Merge the corresponding `m-lab/access` PR to `main`.
2. Re-pin to the merged commit (or a tagged release) here.
3. Mark this PR ready for review and merge.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/260)


handler/handler.go
```diff
@@ -215,13 +225,13 @@ func (c *Client) Nearest(rw http.ResponseWriter, req *http.Request) {
 		}
 	}
 
-	c.handleNearestRequest(rw, req, &result, "nearest")
+	c.handleNearestRequest(rw, req, &result, "nearest", nil)
 }
 
 // handleNearestRequest is a helper that contains the common logic for looking up
 // nearest machines and generating the response. It's used by both Nearest and
 // PriorityNearest handlers.
```
bassosimone, line 233: Please, document that `*IntegrationClaims` could be `nil` and what is the semantics of leaving it nil.

handler/handler.go
```diff
@@ -385,8 +395,20 @@ func (c *Client) PriorityNearest(rw http.ResponseWriter, req *http.Request) {
 		}
 	}
 
+	// Extract key_id claim if present.
+	keyID := ""
+	if keyIDClaim, ok := claims["key_id"]; ok {
+		if v, ok := keyIDClaim.(string); ok {
+			keyID = v
+		}
+	}
```
bassosimone, line 416: I cannot find this line 😿

bassosimone: Also, same question of the other pull request. Is it acceptable that we have the integration ID and not the key ID? What is the case/condition in which this will happen? What is the design principle here underpinning rejecting missing integration ID but allowing missing key ID?

handler/handler.go
```diff
@@ -475,17 +497,17 @@ func (c *Client) checkClientLocation(rw http.ResponseWriter, req *http.Request)
 }
 
 // populateURLs populates each set of URLs using the target configuration.
-func (c *Client) populateURLs(targets []v2.Target, ports static.Ports, exp string, pOpts paramOpts) {
+func (c *Client) populateURLs(targets []v2.Target, ports static.Ports, exp string, pOpts paramOpts, ic *IntegrationClaims) {
```
bassosimone, line 516: Please, document that `ic` may be `nil` and what it means for it to be `nil`.

handler/handler.go
```diff
@@ -475,17 +497,17 @@ func (c *Client) checkClientLocation(rw http.ResponseWriter, req *http.Request)
...
 		params := c.extraParams(target.Machine, i, pOpts)
-		targets[i].URLs = c.getURLs(ports, target.Hostname, token, params)
+		targets[i].URLs = c.getURLs(ports, target.Hostname, tkn, params)
 	}
 }
 
 // getAccessToken allocates a new access token using the given machine name as
 // the intended audience and the subject as the target service.
-func (c *Client) getAccessToken(machine, subject string) string {
+func (c *Client) getAccessToken(machine, subject string, ic *IntegrationClaims) string {
```
bassosimone, line 532: Please, document that `ic` may be `nil` and what it means for it to be `nil`.

bassosimone (commented):
Pre-review with comments that popped up while doing a first read of the diff.

PR #267, opened 2026-06-10
Add per-machine probability override

robertodauria:
Flag to override an individual machine's pick probability by name. Needed for an autojoined machine stuck at `probability: 1`: its registration comes from per-org Datastore metadata (not siteinfo), config is per-org not per-machine, and we can't disable the whole org.

`-probability-override=<machine>=0` (repeatable, also via env). Get the name from `/v2/siteinfo/registrations`. Overridden sites bypass `alwaysPick` so they're filtered even for org/virtual-targeted queries. `go test ./...` and `go vet ./...` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/locate/267)


bassosimone (commented):
@robertodauria I am wondering whether doing this using the command line is going to cause too much friction and I wonder whether a more agile solution could be of loading this information from `Datastore` could allow us to make this functionality easier to control and manage. Happy to chat more about this with you tomorrow!

bassosimone (approved):
To clarify on my latest comment. I have no objections to merging this pull request as is as long as the technical debt is clearly documented as I specified in my previous message.

